PR #1, opened 2018-11-07
package hexyl to archlinux

attenuation:
I have already packaged hexyl to [aur](https://aur.archlinux.org/packages/hexyl/) for archlinux


sharkdp:
Thank you very much!

Issue #4, opened 2018-11-09
Display input offset


sharkdp:
... in hexadecimal, just like `hexdump -C`:
```
00000000  89 50 4e 47 0d 0a 1a 0a  00 00 00 0d 49 48 44 52  |.PNG........IHDR|
00000010  00 00 00 80 00 00 00 44  08 02 00 00 00 c6 25 aa  |.......D......%.|
00000020  3e 00 00 00 c2 49 44 41  54 78 5e ed d4 81 06 c3  |>....IDATx^.....|
00000030  30 14 40 d1 b7 34 dd ff  ff 6f b3 74 56 ea 89 12  |0.@..4...o.tV...|
00000040  6c 28 73 e2 aa 34 49 03  87 d6 fe d8 7b 89 bb 52  |l(s..4I.....{..R|
00000050  8d 3b 87 fe 01 00 80 00  00 10 00 00 02 00 40 00  |.;............@.|
00000060  00 08 00 00 01 00 20 00  00 04 00 80 00 00 10 00  |...... .........|
00000070  00 02 00 40 00 00 08 00  00 01 00 20 00 00 00 d4  |...@....... ....|
00000080  5e 6a 64 4b 94 f5 98 7c  d1 f4 92 5c 5c 3e cf 9c  |^jdK...|...\\>..|
00000090  3f 73 71 58 5f af 8b 79  5b ee 96 b6 47 eb f1 ea  |?sqX_..y[...G...|
000000a0  d1 ce b6 e3 75 3b e6 b9  95 8d c7 ce 03 39 c9 af  |....u;.......9..|
000000b0  c6 33 93 7b 66 37 cf ab  bf f9 c9 2f 08 80 00 00  |.3.{f7...../....|
000000c0  10 00 00 02 00 40 00 00  08 00 00 01 00 20 00 00  |.....@....... ..|
000000d0  04 00 80 00 00 10 00 00  02 00 40 00 00 08 00 00  |..........@.....|
000000e0  01 00 20 00 00 8c 37 db  68 03 20 fb ed 96 65 00  |.. ...7.h. ...e.|
000000f0  00 00 00 49 45 4e 44 ae  42 60 82                 |...IEND.B`.|
000000fb
```

kilobyte:
+1, a near clone of hexdump would be nice.

sharkdp:
Closed via #7 by @PewZ.

PR #7, opened 2018-12-01
display input offset

PewZ:
Hey! Thanks for this awesome tool :)
I added support for displaying input offset at the start of each line. Ref #4 .

This could also be implemented as a command-line flag or something if you'd rather want that.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -91,7 +91,8 @@ impl<'a> Printer<'a> {
     fn header(&mut self) {
         writeln!(
             self.stdout,
-            "┌{}┬{}┬{}┬{}┐",
+            "┌{}┬{}┬{}┬{}┬{}┐",
+            "─".repeat(8),
             "─".repeat(3 * 8 + 1),
             "─".repeat(3 * 8 + 1),
             "─".repeat(8),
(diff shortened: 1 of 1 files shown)


sharkdp:
Awesome, thank you very much for your contribution.

src/main.rs
```diff
@@ -112,7 +114,8 @@ impl<'a> Printer<'a> {
 
     fn print_byte(&mut self, b: u8) -> io::Result<()> {
         if self.idx % 16 == 1 {
-            write!(self.stdout, "│ ");
+            let byte_str = format!("{:08x}", self.idx - 1);
+            write!(self.stdout, "│{}│ ", Byte(0x00).color().paint(byte_str))?;
```
sharkdp: We could use `COLOR_NULL` here instead of `Byte(0x00).color()`. But I'd prefer to add a new constant for this (`COLOR_OFFSET`). It could be the same grey color as for NULL bytes, though.

PewZ: Sounds like a good idea! I've pushed a new fix. Let me know if there's anything else you want me to change.

sharkdp:
Thank you for the updates!

PR #10, opened 2019-01-09
Update README.md

BlackLotus:
Changed yaourt to yay.
Could be pakku, pikaur, aurman or something 
But yaourt is out of date, insecure and not recommended anymore
https://wiki.archlinux.org/index.php/AUR_helpers#Pacman_wrappers


sharkdp:
Thanks! I also switched to `yay`.

PR #21, opened 2019-01-09
Make clear we didn't accidentally forget to handle Results

Byron-TW:
This makes it compile without warnings.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -98,7 +98,7 @@ impl<'a> Printer<'a> {
             "─".repeat(3 * 8 + 1),
             "─".repeat(8),
             "─".repeat(8)
-        );
+        ).ok();
     }
 
     fn footer(&mut self) {
@@ -110,7 +110,7 @@ impl<'a> Printer<'a> {
             "─".repeat(3 * 8 + 1),
             "─".repeat(8),
             "─".repeat(8)
-        );
+        ).ok();
     }
 
     fn print_byte(&mut self, b: u8) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

kballard:
The correct way to do this is to use `let _ = …`. The usage of `….ok();` makes it look like you still forgot to handle the results.

Also, why are they being ignored anyway? Writes earlier in this function use `?` to handle the results; these ones probably should too.

sharkdp:
> The correct way to do this is to use `let _ = …`. The usage of `….ok();` makes it look like you still forgot to handle the results.

I disagree. I often use `….ok()` to signal that I am "okay" with ignoring errors here. I'm explicitly turning `Err(..)` into `None`. Ignoring a `Result<..>` is not okay (throws a warning), ignoring `Option<..>` is fine. That's at least my understanding/interpretation.

> Also, why are they being ignored anyway? Writes earlier in this function use `?` to handle the results; these ones probably should too.

`write!` typically errors if the file handle is already closed. In our case, this signals a broken output pipe. This can happen if the output of `hexyl` is piped into `head 10`, for example. `head` will close its STDIN (our STDOUT) when 10 lines have been printed. The error handling with `?` is here to short-circuit in this case: https://github.com/Byron-TW/hexyl/blob/3a191679d772657714a184333ad6fb809515e8e5/src/main.rs#L234-L237

However, using `.ok()` is also fine on header and footer since we will short-circuit (or exit) shortly afterwards anyway. Using `?` error handling in a tight loop might even incur some overhead due to the conditional.

PR #23, opened 2019-01-10
Replace all uses of format!()

kballard:
Replace all uses of format!()

Previously we were allocating 2 strings per byte, plus another for the offset. Instead we can just write all this directly to stdout.

Also replace the calculation of the ANSI escape code for every single byte. This produces a significant speedup.

I tested this by ensuring the output of dumping a 1.3MB file (the hexyl binary itself) was byte-for-byte identical to without this PR.

Benchmark output:

```
Benchmark #1: ./original_hexyl ./original_hexyl
  Time (mean ± σ):      1.462 s ±  0.005 s    [User: 1.413 s, System: 0.046 s]
  Range (min … max):    1.456 s …  1.469 s
 
Benchmark #2: target/release/hexyl ./original_hexyl
  Time (mean ± σ):     289.6 ms ±   1.5 ms    [User: 243.8 ms, System: 44.1 ms]
  Range (min … max):   288.2 ms … 293.1 ms
 
Summary
  'target/release/hexyl ./original_hexyl' ran
    5.05 ± 0.03 times faster than './original_hexyl ./original_hexyl'
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -78,6 +78,8 @@ struct Printer<'a> {
     idx: usize,
     line: Vec<u8>,
     stdout: StdoutLock<'a>,
+    byte_hex_table: Vec<String>,
+    byte_char_table: Vec<String>,
 }
 
 impl<'a> Printer<'a> {
@@ -86,6 +88,17 @@ impl<'a> Printer<'a> {
             idx: 1,
             line: vec![],
             stdout,
+            byte_hex_table: (0u8..=u8::max_value())
+                .map(|i| format!("{}", Byte(i).color().paint(format!("{:02x} ", i))))
+                .collect(),
+            byte_char_table: (0u8..=u8::max_value())
+                .map(|i| {
+                    format!(
+                        "{}",
+                        Byte(i).color().paint(format!("{}", Byte(i).as_char()))
+                    )
+                })
+                .collect(),
         }
     }
 
@@ -109,12 +122,17 @@ impl<'a> Printer<'a> {
 
     fn print_byte(&mut self, b: u8) -> io::Result<()> {
         if self.idx % 16 == 1 {
-            let offset_str = format!("{:08x}", self.idx - 1);
-            write!(self.stdout, "│{}│ ", COLOR_OFFSET.paint(offset_str))?;
+            let style = COLOR_OFFSET.normal();
+            write!(
+                self.stdout,
+                "│{}{:08x}{}│ ",
+                style.prefix(),
+                self.idx - 1,
+                style.suffix()
+            )?;
         }
 
-        let byte_str = format!("{:02x} ", b);
-        write!(self.stdout, "{}", Byte(b).color().paint(byte_str))?;
+        write!(self.stdout, "{}", self.byte_hex_table[b as usize])?;
         self.line.push(b);
 
         match self.idx % 16 {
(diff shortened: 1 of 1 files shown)


sharkdp:
> Another benefit of using a lookup table is we could stuff the ANSI color escapes in there as well (though in this case we'd definitely want to generate it at runtime), which would avoid having to categorize each byte in order to color it and avoid having to synthesize ANSI escapes for each byte.

That is true, yes. That could be a more significant speedup.

In this case, I'd like to structure the lookup-table login into a new file and struct.

kballard:
Wow, getting rid of the ANSI escape calculation per byte is a *huge* win:

```
Benchmark #1: ./hexyl_lookup_table ./original_hexyl
  Time (mean ± σ):      1.193 s ±  0.005 s    [User: 1.143 s, System: 0.046 s]
  Range (min … max):    1.188 s …  1.206 s
 
Benchmark #2: target/release/hexyl ./original_hexyl
  Time (mean ± σ):     739.4 ms ±   1.7 ms    [User: 691.2 ms, System: 45.8 ms]
  Range (min … max):   737.4 ms … 743.8 ms
 
Summary
  'target/release/hexyl ./original_hexyl' ran
    1.61 ± 0.01 times faster than './hexyl_lookup_table ./original_hexyl'
```

I'll push that momentarily.

sharkdp:
Awesome! Thank you very much.

This looks clean enough to leave it in the main file. And we can still refactor later..

sharkdp:
How much do we gain by using `get_unchecked`? Could we just use `[..]` and get rid of the unsafe block?

kballard:
If we apply the same trick to the char version we get even faster:

```
Benchmark #1: ./original_hexyl ./original_hexyl
  Time (mean ± σ):      1.462 s ±  0.005 s    [User: 1.413 s, System: 0.046 s]
  Range (min … max):    1.456 s …  1.469 s
 
Benchmark #2: target/release/hexyl ./original_hexyl
  Time (mean ± σ):     289.6 ms ±   1.5 ms    [User: 243.8 ms, System: 44.1 ms]
  Range (min … max):   288.2 ms … 293.1 ms
 
Summary
  'target/release/hexyl ./original_hexyl' ran
    5.05 ± 0.03 times faster than './original_hexyl ./original_hexyl'
```

I'll test the non-unsafe version in a minute.

kballard:
The `unsafe` is just noise.

sharkdp:
Fantastic, thank you very much for this real time updates! :smile: 

I'd like to get this merged so I can include it in v0.3.1, which I'd like to release now. Happy to work on even more improvements in other PRs.

kballard:
Sounds good to me. I need to go to bed anyway.

sharkdp:
We're faster than `hexdump -C` now :fireworks: :smile: 

```
▶ hyperfine 'hexyl $(which hexyl)' 'hexdump -C $(which hexyl)'
Benchmark #1: hexyl $(which hexyl)
  Time (mean ± σ):     169.6 ms ±   3.4 ms    [User: 147.7 ms, System: 21.7 ms]
  Range (min … max):   165.4 ms … 179.4 ms    17 runs
 
Benchmark #2: hexdump -C $(which hexyl)
  Time (mean ± σ):     190.0 ms ±   2.5 ms    [User: 188.2 ms, System: 1.7 ms]
  Range (min … max):   186.1 ms … 193.8 ms    15 runs
 
Summary
  'hexyl $(which hexyl)' ran
    1.12 ± 0.03 times faster than 'hexdump -C $(which hexyl)'
```

PR #24, opened 2019-01-10
Get rid of the use of str.repeat()

kballard:
This allocates a string. Instead we can (ab)use format specifiers to do the repetitions for us.

This produces a very mild speedup:

```
Benchmark #1: ./original_hexyl original_hexyl
  Time (mean ± σ):      1.482 s ±  0.016 s    [User: 1.426 s, System: 0.050 s]
  Range (min … max):    1.468 s …  1.516 s
 
Benchmark #2: ./target/release/hexyl original_hexyl
  Time (mean ± σ):      1.473 s ±  0.002 s    [User: 1.423 s, System: 0.046 s]
  Range (min … max):    1.472 s …  1.480 s
 
Summary
  './target/release/hexyl original_hexyl' ran
    1.01 ± 0.01 times faster than './original_hexyl original_hexyl'
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -92,25 +92,19 @@ impl<'a> Printer<'a> {
     fn header(&mut self) {
         writeln!(
             self.stdout,
-            "┌{}┬{}┬{}┬{}┬{}┐",
-            "─".repeat(8),
-            "─".repeat(3 * 8 + 1),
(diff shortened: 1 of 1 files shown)


sharkdp:
That's pretty cool! Thank you very much.

Issue #6, opened 2018-11-15
Unicode dump


kilobyte:
A version of hexdump I wanted to write but never got around to, would do:

 * display Unicode characters on the right side, plus a colored filler (``…``? ``␣``?) for the space not taken (for wcwidth()==1 it's 1 extra character for <U+0800, 2 extra for <=U+FFFF, 3 extra for non-BMP; likewise for wcwidth()==2).  Obviously, a CJK (width 2) character might go over the right edge but as it always takes at least three bytes, there'll be enough space in the next line.

 * display controls using appropriate symbols — Unicode provides a set for this exact task at U+2400. Some specific characters could be better shown using more readable symbols: ``♪`` for 07, ``↵`` for 0a, ``⎋`` for 1b, ``⌫`` for 08, ``⌦`` for 7f, and especially one of for ``⌀`` ``⌾`` ``⍉`` ``␥`` for 00.  You do get a lot of nulls, newlines and escapes in dumps...

 * with an option, display Unicode code points rather than individual bytes on the left side — ie, "U+FFFD" instead of "ef bf bd".

Sounds like your **hexyl** would be a perfect place to implement the above...

kilobyte:
WRT the right side display: currently more than 2/3 of byte values show up as ``.``.  It'd be good to use it for 2e and nothing else.

sharkdp:
Thank you very much for the feedback.

> * display controls using appropriate symbols — Unicode provides a set for this exact task at U+2400. Some specific characters could be better shown using more readable symbols: `♪` for 07, `↵` for 0a, `⎋` for 1b, `⌫` for 08, `⌦` for 7f, and especially one of for `⌀` `⌾` `⍉` `␥` for 00.  You do get a lot of nulls, newlines and escapes in dumps...

I also think this would be really cool! I actually started writing this tool after implementing [this feature](https://github.com/sharkdp/bat/pull/395) in my `bat` tool which does something similar for text-based files.


> * display Unicode characters on the right side, plus a colored filler (`…`? `␣`?) for the space not taken (for wcwidth()==1 it's 1 extra character for <U+0800, 2 extra for <=U+FFFF, 3 extra for non-BMP; likewise for wcwidth()==2).  Obviously, a CJK (width 2) character might go over the right edge but as it always takes at least three bytes, there'll be enough space in the next line.

Obviously this would only work if the file actually contains Unicode text in some form of encoding. So I think this would need to be behind a command-line flag, right?

[...]

PR #30, opened 2019-01-11
Add --color flag to disable ANSI sequences

bennetthardwick:
When printing the output of hexyl to a file (`hexyl file.exe > file.txt`), the `file.txt` file gets polluted by the ANSI escape sequences. Checking whether the application has been launched from inside the terminal or not, allows you to decide whether or not to use them.

I don't have access to a Windows machine, but [atty](https://crates.rs/crates/atty) claims that it has been tested on Windows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,6 +41,7 @@ name = "hexyl"
 version = "0.3.1"
 dependencies = [
  "ansi_term 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "atty 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "clap 2.32.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ version = "0.3.1"
 
 [dependencies]
 ansi_term = "0.11"
+atty = "0.2"
 
 [dependencies.clap]
 version = "2"
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,7 @@
 #[macro_use]
 extern crate clap;
 extern crate ansi_term;
+extern crate atty;
 
 use std::fs::File;
 use std::io::{self, prelude::*, StdoutLock};
@@ -9,6 +10,8 @@ use clap::{App, AppSettings, Arg};
 
 use ansi_term::Colour;
 use ansi_term::Colour::Fixed;
+
+use atty::Stream;
 
 const BUFFER_SIZE: usize = 256;
 
@@ -138,7 +141,7 @@ impl<'a> Printer<'a> {
     fn print_byte(&mut self, b: u8) -> io::Result<()> {
         if self.idx % 16 == 1 {
             let style = COLOR_OFFSET.normal();
-            let byte_index = format!("{:08x}", self.idx -1);
+            let byte_index = format!("{:08x}", self.idx - 1);
             let formatted_string = if self.show_color {
                 format!("{}", style.paint(byte_index))
             } else {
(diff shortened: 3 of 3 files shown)


kilobyte:
Note that this breaks a common use case, of redirecting ``|less -R``.  On the other hand, so do almost all similar programs, so this would be consistent.  On the third hand, overriding isatty() is tedious without one of my, little known, tools (``pipetty``).

roryokane:
> Note that this breaks a common use case, of redirecting `|less -R`.

If color codes were not emitted through pipes by default, that use-case could be restored by requiring a `--color` flag that would force ANSI escape sequences to be emitted.

sharkdp:
@bennetthardwick Thank you very much for your contribution!

> Note that this breaks a common use case, of redirecting `|less -R`.

True. Adding a `--color=auto/always/never` flag could be one option. This is what I did with https://github.com/sharkdp/fd and https://github.com/sharkdp/bat. However, this would require users to type
```
hexyl --color=always … | less -R
```
if they wanted to use a pager. To solve this, we could think about adding a `--pager[=CMD]` flag, similar to how this is done in https://github.com/sharkdp/bat. I don't want to enable paging by default (even with the "quit if one screen" setting of `less`) because that has caused so much pain in `bat`. But I'd be fine with a opt-in feature.

For now, I'd say we merge this and add a `--color` flag afterwards.

kilobyte:
There's also an env var named ``CLICOLOR_FORCE`` that a bunch of programs support.  While I despise its name on aesthethic grounds, it might be good to have it behave same as ``--color=always``.

kballard:
If you don't want color, why use `hexyl`? That's kind of its defining purpose.

sharkdp:
> If you don't want color, why use `hexyl`? That's kind of its defining purpose.

I fully agree. But what if somebody was looking at the colored output and now wants to write that to a file or use `grep` to filter the output? Yes, he could now switch to `hd`/`xxd`, but it's easier to just bring up the last command and pipe it somewhere else.

kballard:
The problem is piping to `less -R` is likely to be a very common use-case, so hurting that use-case in favor of the less common case of piping to a file is problematic.

sharkdp:
Okay, I tend to agree. @bennetthardwick what do you think?

kilobyte:
It depends on how far are you going to extend *hexyl*.  Is it going to be just a simple colorized *hd*?  If so, there's no reason to not use hd when colors are not needed (and the user can use ``ansi2txt`` if for some reason he/she wants *hexyl*).

But if there's a plan for anything that *hd* can't do, then piping definitely should be supported, and in such case you'd want consistency with other programs.  I mean, if for example you'd implement #6, it'd be something useful even without color.

So it's about how many features you plan for.

kballard:
We could add a `--no-color` flag if someone wants to dump to a file without trying to auto-detect color support.

bennetthardwick:
I agree with @kballard, if there's an existing use case that's quite popular it wouldn't be great to break it. `hexyl` could support the `--color` flag and default to `always, which would allow anyone who doesn't use `less -R` to set `--color=auto` as an alias or something. But I think creating an argument that disables the colours is probably the best idea.

sharkdp:
> `hexyl` could support the `--color` flag and default to `always`, which would allow anyone who doesn't use `less -R` to set `--color=auto` as an alias or something. But I think creating an argument that disables the colours is probably the best idea.

That sounds very good to me. If you want to work on this, please amend this PR. If not, feel free to close it and somebody else can work on this. In any case, thank you very much!

bennetthardwick:
Sweet! I've updated the PR to use the `--no-color` flag instead of detecting whether the output is to a terminal.

sharkdp:
Thank you very much! I have updated the PR to use `--color=always/auto/never` (defaulting to `always`) instead of `--no-color`. This way, users can set
``` bash
alias hexyl="hexyl --color=auto"
```
if they want the TTY-detection behavior.

PR #31, opened 2019-01-11
Use `get_matches_safe` instead of `get_matches` for consistent exit point from program

arnavb:
[`get_matches`](https://docs.rs/clap/2.32.0/clap/struct.App.html#method.get_matches) currently displays an error to the user on failure and calls [`process::exit`](https://doc.rust-lang.org/std/process/fn.exit.html). This means that if some sort of error occurs with CLI parsing, then the end of the `main` function is not reached. While this isn't necessarily a problem here, multiple exit points make the code harder to reason about in general. I've replaced `get_matches` with [`get_matches_safe`](https://docs.rs/clap/2.32.0/clap/struct.App.html#method.get_matches_safe), which instead of exiting the program returns a [`ClapResult`](https://docs.rs/clap/2.32.0/clap/type.Result.html). The error is then propagated upwards with the `?` operator, allowing for a consistent exit point on error from the program.

Possible problems with this:
- Calling `--help` and `--version` will currently _always_ be an error (the user won't see an error, but there will be an exit code of 1). I can however check if the error is of kind `ErrorKind::HelpDisplayed` or `ErrorKind::VersionDisplayed` and in those cases exit with a return code of 0. If you feel this is necessary, I will gladly make the appropriate changes to this PR.

Please let me know your thoughts on this and whether I should make any changes.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -258,13 +258,11 @@ fn main() {
     let _ = ansi_term::enable_ansi_support();
 
     let result = run();
-    match result {
-        Err(err) => {
-            // The whitespace is being removed from the right because
-            // clap errors add a newline to the end of the output already
-            eprintln!("Error: {}", format!("{}", err).trim_right());
-            std::process::exit(1);
-        }
-        Ok(()) => {}
+
+    if let Err(err) = result {
+        // The whitespace is being removed from the right because
+        // clap errors add a newline to the end of the output already
+        eprintln!("Error: {}", format!("{}", err).trim_right());
+        std::process::exit(1);
     }
 }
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much.

> * Calling `--help` and `--version` will currently _always_ be an error (the user won't see an error, but there will be an exit code of 1). I can however check if the error is of kind `ErrorKind::HelpDisplayed` or `ErrorKind::VersionDisplayed` and in those cases exit with a return code of 0. If you feel this is necessary, I will gladly make the appropriate changes to this PR.

Yes, we should do this. If you could open another PR for this, that'd be great. Thank you!

PR #33, opened 2019-01-12
Write lines into an in-memory buffer before printing to stdout

kballard:
Running them through an in-memory buffer first produces a 25% speedup (when tested against a 1.3MB file).

Also bump the buffer size to 256 bytes. On my machine this pushes it to a 33% speedup.

```
Benchmark #1: ./hexyl-0.3.1 ./hexyl-0.3.1
  Time (mean ± σ):     277.7 ms ±   1.3 ms    [User: 230.7 ms, System: 45.2 ms]
  Range (min … max):   275.9 ms … 280.8 ms
 
Benchmark #2: target/release/hexyl ./hexyl-0.3.1
  Time (mean ± σ):     184.2 ms ±   1.4 ms    [User: 151.7 ms, System: 30.9 ms]
  Range (min … max):   181.4 ms … 186.0 ms
 
Summary
  'target/release/hexyl ./hexyl-0.3.1' ran
    1.51 ± 0.01 times faster than './hexyl-0.3.1 ./hexyl-0.3.1'
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -76,7 +76,10 @@ impl Byte {
 
 struct Printer<'a> {
     idx: usize,
-    line: Vec<u8>,
+    /// The raw bytes used as input for the current line.
+    raw_line: Vec<u8>,
+    /// The buffered line built with each byte, ready to print to stdout.
+    buffer_line: Vec<u8>,
     stdout: StdoutLock<'a>,
     byte_hex_table: Vec<String>,
     byte_char_table: Vec<String>,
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -76,7 +76,10 @@ impl Byte {
 
 struct Printer<'a> {
     idx: usize,
-    line: Vec<u8>,
+    /// The raw bytes used as input for the current line.
+    raw_line: Vec<u8>,
+    /// The buffered line built with each byte, ready to print to stdout.
+    buffer_line: Vec<u8>,
```
bjorn3: Maybe use https://doc.rust-lang.org/stable/std/io/struct.BufWriter.html instead. (`stdout: io::BufWriter<StdOutLock<'a>>`)

sharkdp:
Thank you very much!

Issue #14, opened 2019-01-09
Show help message when using `hexyl` with no argument.


calio:
Just typing `hexyl` without any argument shows just the header of the output. Better to print nothing or help message.

sharkdp:
Thank you for the feedback.

In this case, `hexyl` is waiting for input from stdin. This is intended in order to allow users to pipe output into `hexyl`, i.e.
```
tail -c 1024 some-file | hexyl
```

ErichDonGubler:
It seems like it'd be pretty rare for somebody to be typing in a hex dump by hand. Maybe printing a warning to `stderr` would be appropriate if a TTY is detected?

sharkdp:
> It seems like it'd be pretty rare for somebody to be typing in a hex dump by hand. Maybe printing a warning to `stderr` would be appropriate if a TTY is detected?

That's probably correct and interactive TTY detection on STDIN would be a way to fix this, yes. Are there any other tools that do something like this, i.e. disable input from STDIN if its from a terminal. `hexdump`, `xxd` and all other Unix tools I can think of do not have this behavior.

ErichDonGubler:
@sharkdp: I don't even necessarily know about disabling input if you print a warning, but that's an interesting idea. AFAIK, Unix is the only set of *common* terminal tools that deal with hex dumps. Windows' Powershell has [`Format-Hex`](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/format-hex?view=powershell-6), but it's fairly hard to get it to use `stdin` directly anyway.

sharkdp:
I think I'd like to just keep this as is. Maybe someone wants to copy & paste some data into the terminal. For example, copy & paste a Unicode string to see its UTF-8 representation with `hexyl`.

Let me know if somebody has strong opinions on this, so I can reopen the ticket.

kilobyte:
+1 — the old Unix design guideline for command-line tools was/is "no news is good news".

PR #42, opened 2019-01-17
Show help if no file / stream is provided

bennetthardwick:
Currently, running `hexyl` without an input stream or file will show the start of the output and then stop without exiting. This PR replaces that behavior and instead shows the help menu and closes with exit code 1.


kilobyte:
This has already been discussed and rejected in #14

Why would that be a problem?  It's how most Unix filter-type tools work if they can reasonably work on plain text input.  So does for example ``hd`` for which ``hexyl`` is a replacement.

Issue #16, opened 2019-01-09
Options for reading between two offsets


ErichDonGubler:
The `-n/--length` flag is a great feature -- good especially for quickly checking if file headers match something. Another use case I can think of for limited output is inspecting, say, an entire block of some data from a file that's been dumped from a disk. Let's say I'm using Linux and reading the first block of a disk somewhere to determine its contents manually:

```
$ hexyl -n 512 "$disk_dump"
// Some output here...
```

I read the output and discover that there's an MBR at the beginning, with the first partition starting at logical block address 1. Sweet, let's mosey on over to 0x200 and read another block. I could implement this by using `dd`:

```
$ input_file="_viminfo" block_size=512 block_num=1
$ dd bs="$block_size" status=none skip="$block_num" count="$block_size" if="$input_file" \
    | hexyl
```

...but there's two issues with this:

* `dd` isn't usually available on Windows machines!
* Because we're using `stdin`, we can't get the correct set of offsets -- the "file" starts at 0x0 regardless of what parameters I gave `dd`.

Perhaps something like this spitballed set of options might help:

```
$ hexyl \
    --start 512      \ # Could also be written as 
                     \ #
                     \ # Could be bikeshed to `--begin`?
                     \
    --length 512     \
                     \
                     \ # One could use an end offset instead of a length:
    # --end 1024     \ # Could also be written as `-e 0x300
```

Having something similar to `bat`'s `--range` could also be really handy, especially when combined with relative offsets (positive and negative):

```sh
$ hexyl --range 512:+512 # same as using `--skip 512 --length 512`
$ hexyl --range=-512: # read the last block
$ hexyl --block-size 4096 -1block: # like above, but use the block unit
$ hexyl --range 32:-32 # cut out a common header and footer for the input stream we don't care about
```

I would be more than happy to push implementation of this, since I've great personal interest in allowing more of my reverse engineering flow in the terminal. Let me know if you want me to hack on it!

### Unresolved questions

* ~~Adding several more arguments that accept offsets/sizes might add pressure to create a system of units/radices a la `xxd` for the appropriate arguments. Where should the line be drawn in terms of what this project is willing to support?~~ We've added support for `xb` and `xib` with #44. That's as far as we've decided to go right now.

ErichDonGubler:
Another idea is to add something to change the initial "offset" of printing -- to reuse the `dd` example:

```
$ input_file="_viminfo" block_size=512 block_num=1
[...]

PR #43, opened 2019-01-17
Options for reading between two offsets

Regentag:
new option `-s'

hexyl_new

related issue : #16

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,7 +5,7 @@ extern crate atty;
 extern crate ctrlc;
 
 use std::fs::File;
-use std::io::{self, prelude::*, StdoutLock};
+use std::io::{self, prelude::*, SeekFrom, StdoutLock};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
@@ -83,7 +83,7 @@ impl Byte {
 struct Printer<'a> {
     idx: usize,
     /// The raw bytes used as input for the current line.
-    raw_line: Vec<u8>,
+    raw_line: Vec<Option<u8>>,
     /// The buffered line built with each byte, ready to print to stdout.
     buffer_line: Vec<u8>,
     stdout: StdoutLock<'a>,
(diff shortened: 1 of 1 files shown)


sharkdp:

I will take a closer look soon.
/cc @ErichDonGubler

src/main.rs
```diff
@@ -251,8 +270,29 @@ fn run() -> Result<(), Box<::std::error::Error>> {
...
+    let mut print_offset = false;
+    let print_idx = start_offset - (start_offset % 16);
+
     let mut reader: Box<dyn Read> = match matches.value_of("file") {
-        Some(filename) => Box::new(File::open(filename)?),
+        Some(filename) => {
+            let mut f = File::open(filename)?;
+
+            // apply start offset
+            if start_offset > 0 {
```
ErichDonGubler: Is there a reason that this is limited to files? I understand that `Stdin` doesn't implement `Seek`, but surely it'd be fairly easy to just ignore input from `StdinLock` by using [`BufRead`](https://doc.rust-lang.org/std/io/trait.BufRead.html)'s `fill_buf` and `consume`, 

One defense I can see for this is that it's not immediately obvious (to me) how to make that robust. See discussion below about errors for my thoughts on this.

src/main.rs
```diff
@@ -251,8 +270,29 @@ fn run() -> Result<(), Box<::std::error::Error>> {
...
+                f.seek(SeekFrom::Start(start_offset))
```
ErichDonGubler: So, question: is this really an error if, say, the file simply wasn't as big as the specified offset? I think that this is a good default behavior, but I'm not sure about the case where the file size simply doesn't match.

This applies to the `stdin` discussion above since it's entirely possible to detect if there are any remaining bytes in `stdin`.

sharkdp:
Happy to review this once the questions by @ErichDonGubler are answered.

ErichDonGubler:
Just pinging @Regentag to take a look at the comments made so far -- excited to get your stuff through, just wanted to have some discussion! :)

keplersj:
Any update on this? I'd love to use this feature @Regentag

Issue #46, opened 2019-01-20
Please use (or support) -c for byte count


FauxFaux:
`hexyl -n 100` is, in my mind, replacing `head -c 100 | xxd`. If it is directly replacing `xxd`, `xxd` calls this option `-l` (`length`, which I had to look up, because I use `head`).

Please support `-c`, with the same behaviour as the current `-n`, and/or improve the error output to remind users of the correct option:

```
% hexyl -c 257 /dev/urandom
error: Found argument '-c' which wasn't expected, or isn't valid in this context

USAGE:
    hexyl [file]

For more information try --help
```

Annoyingly, `xxd` uses `-c` for `cols`.

sharkdp:
I used `-n` for `--length` because that is used in `hexdump`. I'm okay with adding an additional alias `-c`.

selfup:
I made a small PR, but multiple shortnames does not seem to be an option for `clap`.

I made some suggestions, but otherwise `--c` would be the easiest compromise 🙏

sharkdp:
@selfup Thanks for looking into this. That's a pitty. Maybe this is something that should be requested in `clap`.

> I made some suggestions, but otherwise `--c` would be the easiest compromise

I'm not a big fan of adding `--c`, to be honest. Long `--` options should be used for descriptive names.

There is a possibility to hide options in `clap`, so I guess we could go with your other idea and just add a second option which could be hidden.

selfup:
Yea hiding the `--c` so we can get a `-c` would be interesting. If I have time I'll see what can be done about an upstream to clap. Otherwise I will attempt to hide the long "fakename"

selfup:
Ok updated my PR: https://github.com/sharkdp/hexyl/pull/48#issuecomment-460449756

selfup:
Let me know if those PRs resolved the issue @FauxFaux :smile:

PR #48, opened 2019-02-03
Add -c short and hide values for --bytes long

selfup:
# c alias for length and -n

References #46

### Usage

```bash
hexyl doc/logo.svg --c 100
```

### Caveats

So this is interesting. There doesn't seem to be a way to pass in multiple shortnames.

You also can't define the same `Arg::with_name("same_name")` twice.

_potentially seeing if clap would be OK with adding or getting a PR for multiple shortnames would be another option as well_

So the closest thing I could get was adding an alias which still means `--` instead of `-`.

***

Now one thing that could be done would be to add a new Arg block but that feels repetitive?

```rust
// support both `--c` as well as `-c`
.arg(
    Arg::with_name("c")
        .short("c")
        .long("c")
        .takes_value(true)
        .value_name("N")
        .help("Read only N bytes from the input"),
)

...

// add new conditional to do the same job as `value_of("length")`
if let Some(count) = matches.value_of("c").and_then(|c| c.parse::<u64>().ok()) {
    reader = Box::new(reader.take(count));
}
```

https://docs.rs/clap/2.20.0/clap/struct.Arg.html#short

### If adding a new `--c` / `-c`

The diff would look like (from current commit in PR)

```diff
--- a/src/main.rs
+++ b/src/main.rs
@@ -228,7 +228,6 @@ fn run() -> Result<(), Box<::std::error::Error>> {
         .arg(Arg::with_name("file").help("File to display"))
         .arg(
             Arg::with_name("length")
-                .alias("c")
                 .short("n")
                 .long("length")
                 .takes_value(true)
@@ -236,6 +235,14 @@ fn run() -> Result<(), Box<::std::error::Error>> {
                 .help("Read only N bytes from the input"),
         )
         .arg(
+            Arg::with_name("c")
+                .short("c")
+                .long("c")
+                .takes_value(true)
+                .value_name("N")
+                .help("Read only N bytes from the input"),
+        )
+        .arg(
             Arg::with_name("color")
                 .long("color")
                 .takes_value(true)
@@ -264,6 +271,10 @@ fn run() -> Result<(), Box<::std::error::Error>> {
         reader = Box::new(reader.take(length));
     }

+    if let Some(count) = matches.value_of("c").and_then(|c| c.parse::<u64>().ok()) {
+        reader = Box::new(reader.take(count));
+    }
+
     let show_color = match matches.value_of("color") {
         Some("never") => false,
         Some("auto") => atty::is(Stream::Stdout),
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -228,6 +228,7 @@ fn run() -> Result<(), Box<::std::error::Error>> {
         .arg(Arg::with_name("file").help("File to display"))
         .arg(
             Arg::with_name("length")
+                .alias("c")
                 .short("n")
                 .long("length")
                 .takes_value(true)
--- a/src/main.rs
+++ b/src/main.rs
@@ -235,9 +235,9 @@ fn run() -> Result<(), Box<::std::error::Error>> {
                 .help("Read only N bytes from the input"),
         )
         .arg(
-            Arg::with_name("c")
+            Arg::with_name("bytes")
                 .short("c")
-                .long("c")
+                .long("bytes")
                 .takes_value(true)
                 .hidden(true),
         )
@@ -263,14 +263,9 @@ fn run() -> Result<(), Box<::std::error::Error>> {
         None => Box::new(stdin.lock()),
     };
 
-    if let Some(length) = matches
-        .value_of("length")
-        .and_then(|n| n.parse::<u64>().ok())
-    {
-        reader = Box::new(reader.take(length));
-    }
+    let length_arg = matches.value_of("length").or(matches.value_of("bytes"));
 
-    if let Some(length) = matches.value_of("c").and_then(|n| n.parse::<u64>().ok()) {
+    if let Some(length) = length_arg.and_then(|n| n.parse::<u64>().ok()) {
         reader = Box::new(reader.take(length));
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -228,12 +228,18 @@ fn run() -> Result<(), Box<::std::error::Error>> {
         .arg(Arg::with_name("file").help("File to display"))
         .arg(
             Arg::with_name("length")
-                .alias("c")
                 .short("n")
                 .long("length")
                 .takes_value(true)
                 .value_name("N")
                 .help("Read only N bytes from the input"),
+        )
+        .arg(
+            Arg::with_name("c")
+                .short("c")
+                .long("c")
+                .takes_value(true)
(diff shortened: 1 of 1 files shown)


selfup:
As discussed @sharkdp I have added `--c` and `-c` and ensured the values are hidden.

We could always change `--c` to `--count` which seems like a viable alias for `--length` in this context.

However this is the path of least resistance! 🎉 

Let me know what you think

src/main.rs
```diff
@@ -263,6 +270,10 @@ fn run() -> Result<(), Box<::std::error::Error>> {
         reader = Box::new(reader.take(length));
     }
 
+    if let Some(length) = matches.value_of("c").and_then(|n| n.parse::<u64>().ok()) {
```
sharkdp: In order not to repeat the parsing and `Box::new` logic, you could use something like this:
``` rust
let length_arg = matches.value_of("length").or(matches.value_of("c"));

if let Some(length) = length_arg.and_then(...) {
    // ...
}
```

src/main.rs
```diff
@@ -234,6 +234,13 @@ fn run() -> Result<(), Box<::std::error::Error>> {
                 .value_name("N")
                 .help("Read only N bytes from the input"),
         )
+        .arg(
+            Arg::with_name("c")
```
sharkdp: What about using `bytes` as the name for this argument (and the `long(..)` name) in order to be consistent with what `head` and `tail` do (as referenced in the original ticket)?

src/main.rs
selfup: Yea looks cleaner, will implement

src/main.rs
selfup: Ok, will change --c to --bytes. Also curious, we are still going to keep this hidden right? Should we add it to the README or just keep it on the secret menu?

selfup:
Ok changed `--c` to `--bytes` with a short or `-c`.

Still hidden from list of commands.

`--length`/`-n`/`-bytes`/`-c` all behave the same now

hexylbyteslength

sharkdp:
Awesome, thank you very much.

sharkdp:
> Still hidden from list of commands (maybe it can be shown now that it has valuable meaning?)

Sorry. Only saw that after merging.

Yes, we could also make it visible and have the help text say something like "An alias for -n/--length".

selfup:
I'll make a new PR to make it visible

Issue #47, opened 2019-01-26
Please don't print the header before input is seen


FauxFaux:
`hexyl` appears to print the "header" (the top of the box) too soon.

If you have an app that thinks before printing output (e.g. `gpg`), it is unnecessarily ugly.

Something like this:

```
% (echo Er... >&2; sleep 0.2; \
 echo Thinking... >&2; sleep 0.2; \
 echo Um... >&2; sleep 0.2; \
 echo Output, finally; \
 echo 'Done!' >&2) | hexyl
```

Typically (it's racy!) looks like:

```
Er...
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
Thinking...
Um...
Done!
│00000000│ 4f 75 74 70 75 74 2c 20 ┊ 66 69 6e 61 6c 6c 79 0a │Output, ┊finally_│
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
```

I'd rather it showed:

```
Er...
Thinking...
Um...
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 4f 75 74 70 75 74 2c 20 ┊ 66 69 6e 61 6c 6c 79 0a │Output, ┊finally_│
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
Done!
```

Gaming the thinking is probably easy. Gaming the `Done!` is probably impossible. (`gpg` doesn't print the done.)

sharkdp:
Sounds good to me.

sharkdp:
Fixed by @mziter in #51.

PR #49, opened 2019-02-04
Fixes issue with broken-up stdout output

mziter:
Fixes issue sharkdp/hexyl#47 by wrapping StdoutLock in BufWriter. This will make sure output is written in chunks, which should lessen the annoyance when other processes write to stdout during the same window as hexyl.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,7 +5,7 @@ extern crate atty;
 extern crate ctrlc;
 
 use std::fs::File;
-use std::io::{self, prelude::*, StdoutLock};
+use std::io::{self, prelude::*, BufWriter, StdoutLock};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
@@ -86,7 +86,7 @@ struct Printer<'a> {
     raw_line: Vec<u8>,
     /// The buffered line built with each byte, ready to print to stdout.
     buffer_line: Vec<u8>,
-    stdout: StdoutLock<'a>,
+    buffer_stdout: BufWriter<StdoutLock<'a>>,
     show_color: bool,
     byte_hex_table: Vec<String>,
     byte_char_table: Vec<String>,
@@ -98,7 +98,7 @@ impl<'a> Printer<'a> {
             idx: 1,
             raw_line: vec![],
             buffer_line: vec![],
-            stdout,
+            buffer_stdout: BufWriter::new(stdout),
             show_color,
             byte_hex_table: (0u8..=u8::max_value())
                 .map(|i| {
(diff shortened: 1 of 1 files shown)


selfup:
Nice! Adding screenshots to show that the `Done!` case will as expected be quite impossible :joy: 

screenshot from 2019-02-04 06-50-28

***

```bash
hexyl (master) $ (echo Er... >&2; sleep 0.2;  echo Thinking... >&2; sleep 0.2;  echo Um... >&2; sleep 0.2;  echo Output, finally;  echo 'Done!' >&2) | target/release/hexyl 
Er...
Thinking...
Um...
Done!
│00000000│ 4f 75 74 70 75 74 2c 20 ┊ 66 69 6e 61 6c 6c 79 0a │Output, ┊finally_│
```

sharkdp:
@mziter Thank you very much for working on this!

Is there any chance that this could cause unwanted behavior in some cases? What if `hexyl` is reading input from STDIN (from the keyboard)? Would the output always be present when the user submits a new line?

Couldn't we just wait with printing the header until we actually write the first line of (actual) output?

mziter:
@sharkdp 
Glad to help. I am new to Rust and open-source work.  I believe I tested the STDIN scenario you describe and the buffer is still flushed, but I will check again now.
The reason I chose to go the buffer route is that I believe the issue really isn't just the header row, but more general. If we expanded FauxFaux's script in his submittted issue, I think we would have seen a lot of other lines broken up as well. When any other processess write to STDOUT they will "interrupt" our pretty output. Using the buffer at least large chunks will get written together. I may try to expand FauxFaux's script to illustrate my point.

sharkdp:
Thank you for the detailed response.

I don't really care too much about the interactive-input scenario, but I'm concerned that there might be a output latency in general, when we introduce the buffered writer.

My question would be: do we really have to support a scenario where a second process is printing to the terminal at the same time?! What would be a use case for this? I see the `( echo Er.. >&2; sleep ...` snippet more like a way to demonstrate that the header is printed early (which could be fixed).

mziter:
I just saw the buffer as a simple way to solve the solution and other similiar problems. I am new to Rust and system languages in general and appreciate your feedback.

I just did a simple benchmark on a large .png file and I do see a significant performance decrease using the extra buffer, so you were obviously right to be concerned. I can do as you suggested and handle just the header row.

I am going to be traveling for most of today, so I may not get to it until tomorrow. Two questions:i
* What is best way to ignore these changes I already made? Cancel the PR and create a new one?
* For implementing change for just the header row, would a bool on the printer be best option to see if header has been printed yet or not?

sharkdp:
> I can do as you suggested and handle just the header row.

 That sounds great, thank you.

> * What is best way to ignore these changes I already made? Cancel the PR and create a new one?

That's probably the easiest way. You can also force-push only the new commits on your branch, if you like.

> * For implementing change for just the header row, would a bool on the printer be best option to see if header has been printed yet or not?

Sounds like a good idea to me, yes.

mziter:
Opening a new pull request with solutioning discussed above.

mziter:
Opening new pull request #51

PR #51, opened 2019-02-06
Fixes issue with header printing before data in some scenarios

mziter:
Fixes issue #47 where header can be printed before the first row of data in some scenarioes.
Attached is before (separate header) and after fix.
hexyl_header_fix


mziter:
Updated pull request because I didnt adjust boolean after printed header. Now prints multiple lines without any issues.

sharkdp:
Thank you very much!

Issue #17, opened 2019-01-09
Use a different vertical separator character 


lzybkr:
On Windows when I use the Fira Code font in Alacritty, the character `┊` is not available so it looks like this:

![image](https://user-images.githubusercontent.com/2148248/50923118-04d37300-1401-11e9-952d-47c80137fcc8.png)

Fira Code seems like a popular font, so it would be nice to use a different character.

sharkdp:
Hm... I also use Fira Code in my terminal (on Linux) and do not have any problems with that symbol.

lzybkr:
Weird - it does seem to be the font though - this is what I see in the Font Settings control panel (and this also shows I have what seems to be the latest version of the font):

![image](https://user-images.githubusercontent.com/2148248/50924778-64338200-1405-11e9-8087-2948da774e40.png)

sharkdp:
Does this also appear with other fonts?

sharkdp:
Does this also appear with other fonts?

roryokane:
Here are some more data points for the terminal emulators **Terminal** and **iTerm2** on **macOS**. In all of the tested fonts, the character is displayed. In some fonts its alignment is wrong, in others it is correct. The choice of terminal emulator does not make a difference to whether the character is correctly aligned.

I draw no conclusion from this data on whether the character should be changed.

# Fonts in which the character is misaligned

## 16pt Consolas
### Terminal
<img width="842" alt="Terminal 16pt Consolas" src="https://user-images.githubusercontent.com/79168/51422933-c585ee80-1b85-11e9-901d-bf1cdc6c9f66.png">

### iTerm2
<img width="842" alt="iTerm2 16pt Consolas" src="https://user-images.githubusercontent.com/79168/51422936-d33b7400-1b85-11e9-8d5e-ad63548118b4.png">

## 16pt Fira Code
### Terminal
<img width="922" alt="Terminal 16pt Fira Code" src="https://user-images.githubusercontent.com/79168/51422978-8d32e000-1b86-11e9-84aa-07ba9b4005b5.png">

### iTerm2
<img width="922" alt="iTerm2 16pt Fira Code" src="https://user-images.githubusercontent.com/79168/51422980-915efd80-1b86-11e9-87ce-e4db335d44af.png">

## 18pt Ubuntu Mono
### Terminal
[...]

PR #54, opened 2019-03-05
Implement various --border styles

dmke:
This might help to remedy display issues described in #17.

Please note, I'm not a Rust developer, so there might be better ways to solve this. In particular, I'm not a fan of the various `match`es on the `border_style` field.

Here's a screenshot of how this looks in my terminal:

image

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -80,6 +80,51 @@ impl Byte {
     }
 }
 
+enum BorderStyle {
+    Unicode,
+    Ascii,
+    None,
+}
+
+impl BorderStyle {
+    /// returns, in order, the left corner, horizontal line, column
+    /// seperator and right corner for the header
+    fn header_elems(&self) -> Option<(char, char, char, char)> {
+        match self {
+            BorderStyle::Unicode => Some(('┌', '─', '┬', '┐')),
+            BorderStyle::Ascii   => Some(('+', '-', '+', '+')),
+            BorderStyle::None    => None,
+        }
+    }
+
+    /// returns, in order, the left corner, horizontal line, column
+    /// seperator and right corner for the footer
+    fn footer_elems(&self) -> Option<(char, char, char, char)> {
+        match self {
+            BorderStyle::Unicode => Some(('└', '─', '┴', '┘')),
+            BorderStyle::Ascii   => Some(('+', '-', '+', '+')),
+            BorderStyle::None    => None,
+        }
+    }
+
+    fn outer_sep(&self) -> char {
+        match self {
+            BorderStyle::Unicode => '│',
+            BorderStyle::Ascii   => '|',
+            BorderStyle::None    => ' ',
+        }
+    }
+
+    fn inner_sep(&self) -> char {
+        match self {
+            BorderStyle::Unicode => '┊',
+            BorderStyle::Ascii   => '|',
+            BorderStyle::None    => ' ',
+        }
+    }
+}
+
+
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -126,20 +134,36 @@ impl<'a> Printer<'a> {
...
+                self.stdout,
+                "┌{0:─<8}┬{0:─<25}┬{0:─<25}┬{0:─<8}┬{0:─<8}┐",
+                ""
+            ),
+            BorderStyle::Ascii => writeln!(
+                self.stdout,
+                "+{0:-<8}+{0:-<25}+{0:-<25}+{0:-<8}+{0:-<8}+",
+                ""
+            ),
+            _ => Ok(())
```
bjorn3: Could you replace `_` here and at other places to `BorderStyle::None`? That will make it harder to forget to change one match when a new border style is introduced.

dmke: Sure, will do.

sharkdp:
I didn't have time to look at the code changes, but I have to say that I really like the style without any border. I tend to think that we should make this the default. What do others think?

dmke:
Currently, without a border, this puts both a leading and a trailing space character on each line. I will try to eliminate that, but I might need some guidance later. Let me come up with the implementation first :-)

src/main.rs
dmke: Done.

dmke:
I've now spend a good amount of time trying to understand the borrow checker. My initial plan was to move some details into something like

```rust
impl BorderStyle {
    // Returns left corner, colum separator, horizontal line and right corner.
    fn header_elems(self) -> Option<(char, char, char, char)> {
        match self {
            BorderStyle::Unicode => Some(('┌', '┬', '─', '┐')),
            BorderStyle::Ascii   => Some(('+', '+', '-', '+')),
            BorderStyle::None    => None,
        }
    }
}
```

and condense `Printer.header` to

```rust
fn header(&mut self) {
    if let Some((l,c,h,r)) = self.border_style.header_elems() {
        writeln!(self.stdout, "...", ...).ok()
    }
}
```

but `self` (the Printer) is in a borrowed context. Moving the `if let` into `Printer`'s sole consumer (the `print_byte` function) would (a) not help either, and (b) make that function even more complex.

Another approach would be to move the `writeln!` into a `BorderStyle.print_header()` function, but that'd need to borrow the Printer's `self.stdout`, so that wouldn't work, either.

Do you have suggestions, on how to make this a bit more DRY?

sharkdp:
I can try to look into this, if you could provide the full code of what you are trying to do.

Would this worl?
``` rust
fn header(&mut self) {
    let header_elems = self.border_style.header_elems();
    if let Some((l,c,h,r)) = header_elems {
        writeln!(self.stdout, "...", ...).ok()
    }
}
```

dmke:
> Would this work?

No, that was not it. I've found a solution, though, by allowing `self.border_style` to be borrowed in the `BorderStyle.header_elems` function (are those the right terms? :-D):

```diff
-fn header_elems(self) -> Option<(char, char, char, char)> {
+fn header_elems(&self) -> Option<(char, char, char, char)> {
```

I've updated the commit (again).

selfup:
Just chiming in. The border none is neat for sure 👍

sharkdp:
@dmke Sorry for the delay. I have a few suggestions, but I think it's easier if I just merge this as is and integrate a changes myself. Thank you very much!

PR #55, opened 2019-03-09
Set --color=auto by default

ignatenkobrain:


sharkdp:
See #30 for why we did not set the default to `auto`.

Issue #53, opened 2019-02-25
Squeeze zero rows


moshohayeb:
A useful feature hexdump is the ability to compress contigous rows of 0s with a *

dd if=/dev/zero of=swapfile bs=1024k count=1
mkswap swapfile
hexdump swapfile

0000000 0000 0000 0000 0000 0000 0000 0000 0000
*
0000400 0001 0000 00ff 0000 0000 0000 5dd8 a838
0000410 6fe7 d247 cc9b 6af5 a41d 50da 0000 0000
0000420 0000 0000 0000 0000 0000 0000 0000 0000
*
0000ff0 0000 0000 0000 5753 5041 5053 4341 3245
0001000 0000 0000 0000 0000 0000 0000 0000 0000
*
0100000

I was wondering if this is something hexyl can implement

sharkdp:
Thank you for the feedback.

I will personally not have the time to implement this, but that sounds like a useful feature!

sharkdp:
Implemented in #59 by @awidegreen.

PR #59, opened 2019-03-30
Adds squeeze support

awidegreen:
The squeeze feature is implement as for `hexdump`. Reoccurring lines will
be squeezed together and visualized with an asterisk. Further `hexyl`
received an additional command line argument `-v|--no-squeezing` to
disable squeezing. Squeezing is enabled by default.

If the squeezing goes until the end of the input data, an empty line
with address next to last line will be printed to indicate the highest
address (again similar to `hexdump`).

Implements #53 except the compressing applies to all possible values, 
not just `0`.

Some example outputs:
image
image
image

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,6 +4,8 @@ extern crate ansi_term;
 extern crate atty;
 extern crate ctrlc;
 
+mod squeezer;
+
 use std::fs::File;
 use std::io::{self, prelude::*, StdoutLock};
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -15,6 +17,7 @@ use ansi_term::Color;
 use ansi_term::Color::Fixed;
 
 use atty::Stream;
+use squeezer::{SqueezeAction, Squeezer};
 
 const BUFFER_SIZE: usize = 256;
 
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you very much for working on this! I fixed some of the conflicts, but then made some changes again. It would be great if you could fix the remaining conflicts and update the branch such that it respects the changes in #54, i.e. uses the non-hardcoded border elements.

If you don't find the time to update the branch, let us know.

src/main.rs
```diff
@@ -341,6 +380,16 @@ fn run() -> Result<(), Box<::std::error::Error>> {
                 .value_name("N")
                 .help("An alias for -n/--length"),
         )
+        .arg(
+            Arg::with_name("nosqueeze")
+                .short("v")
+                .long("no-squeeze")
```
sharkdp: Can we be consistent with `hexdump` and call this `--no-squeezing`?

awidegreen: certainly!

awidegreen:
Rebased and changed command line option to `-v | --no-squeezing`

sharkdp:
Thank you very much!

PR #68, opened 2019-07-17
crate: add library

### LINKED ISSUE #67: Provide library as well as binary
I'm interested in using `hexyl` as an hex viewer internally in another project I'm working on.

It would be great to have access to `hexyl` as a library, and to have it operate on generic `Read` and `Write` traits, rather than being tightly coupled to `stdin`/`stdout`.

tommilligan:
Closes #67. 

- moves the majority of `main.rs` to `lib.rs`
  - updates `Printer` to accept any `Write`r
  - factors out `mainloop` as a new method `Printer.print_all`, which accepts any `Read`er
  - adds tests for `print_all`, as we can now read and write directly to buffers in memory
- adds `bin/hexyl.rs`

This PR does not change any public functionality.


sharkdp:
Awesome work, thank you very much! I should have done this from the beginning, but never got around to it.

Let me know if you would like this to be released as a new version.

tommilligan:
Thanks, it's a really cute project! I'm already installing it from git so don't need a release anytime soon.

You may want to double check the API before you release it and have to support it...

PR #70, opened 2019-08-26
Optimize Test Organization

liangyongrui:


sharkdp:
Thank you for your contribution.

Could you tell us what the benefits would be?

liangyongrui:
> Thank you for your contribution.
> 
> Could you tell us what the benefits would be?

I found that the previous unit tests were all public apis.

I agree with the "[The Rust Programming Language](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests)":

> In Rust, integration tests are entirely external to your library. They use your library in the same way any other code would, which means they can only call functions that are part of your library’s public API. Their purpose is to test whether many parts of your library work together correctly. Units of code that work correctly on their own could have problems when integrated, so test coverage of the integrated code is important as well. To create integration tests, you first need a tests directory.

sharkdp:
That is true for integration tests. But not for unit tests. https://doc.rust-lang.org/book/ch11-03-test-organization.html#unit-tests

Issue #69, opened 2019-08-12
Additional space on last line


rfuest:
If the last line of data contains less than 8 bytes, an additional space is written after the `outer_sep`:
https://github.com/sharkdp/hexyl/blob/175847787dd44cfa9536cb1a0980e81b450a587e/src/lib.rs#L340-L349

```bash
$ echo -n "12345678" | hexyl --color never --border ascii | wc -L
80

$ echo -n "1234567" | hexyl --color never --border ascii | wc -L
81
```
This causes an empty line before the footer if the the terminal is exactly 80 characters wide. Is this behavior intended or a bug, because the `short_input_passes` test also contains the tailing space?

sharkdp:
That looks like a bug to me and should be fixed.

Thank you for reporting this!

PR #71, opened 2019-10-01
Remove additional space found on short input

nalshihabi:
Addresses #69 

So looking at the case it seemed that in the formatting code there was just a space included at the end extra, so I changed for it to be no extra space. This required also that the test `short_input_passes` also be changed to reflect that there should be no extra space (otherwise it would currently fail without the extra space).


sharkdp:
Looks good, thank you very much!

PR #75, opened 2019-10-11
Option for reading between two offsets

MaxJohansen:
I have tried to implement the syntax suggested in the discussion on issue #16.

```
$ hexyl --range 512:1024 # Means "start at 512th byte, read until byte 1024 (512 bytes)"
$ hexyl --range 512:+512 # Means "start at 512th byte, read 512 bytes"

$ hexyl --range +512:    # Means "start at 512th byte".
```

I did not add support for negative offsets simply because I do not see a simple way to implement it for `Stdin`. If required, I suppose it could be implemented with a buffering solution that reads `M - N` bytes into a buffer until it reaches the end of `Stdin` before passing it to a reader.

My solution differs from #43 by not modifying the Printer itself, merely the Reader that it receives. This means that it works just as well for `Stdin` as for a file. A consequence of this is that the bytes are no longer automatically "aligned" in the output, but that just might be what you want in this situation.

```
$ hexyl -r :+0x20 hexyl
│00000000│ 7f 45 4c 46 02 01 01 00 ┊ 00 00 00 00 00 00 00 00 │•ELF•••0┊00000000│
│00000010│ 03 00 3e 00 01 00 00 00 ┊ 30 d1 02 00 00 00 00 00 │•0>0•000┊0×•00000│

$ hexyl -r 0x3:+0x20 hexyl
│00000003│ 46 02 01 01 00 00 00 00 ┊ 00 00 00 00 00 03 00 3e │F•••0000┊00000•0>│
│00000013│ 00 01 00 00 00 30 d1 02 ┊ 00 00 00 00 00 40 00 00 │0•0000×•┊00000@00│
```

This is my first contribution of Rust code and I welcome suggestions that can help me improve! :)

### DIFF
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -252,7 +252,7 @@ mod tests {
     #[test]
     fn parse_empty_range() {
         let (offset, bytes_to_read) = parse_range(":").expect("Not allowed to fail test.");
-        assert_eq!(offset, u64::min_value());
+        assert_eq!(offset, 0);
         assert_eq!(bytes_to_read, u64::max_value());
     }
 
@@ -270,11 +270,11 @@ mod tests {
     #[test]
     fn parse_ending_offset() {
         let (offset, bytes_to_read) = parse_range(":0x200").expect("Not allowed to fail test.");
-        assert_eq!(offset, u64::min_value());
+        assert_eq!(offset, 0);
         assert_eq!(bytes_to_read, 512);
 
         let (offset, bytes_to_read) = parse_range(":+512").expect("Not allowed to fail test.");
-        assert_eq!(offset, u64::min_value());
+        assert_eq!(offset, 0);
         assert_eq!(bytes_to_read, 512);
 
         let (offset, bytes_to_read) = parse_range("512:512").expect("Not allowed to fail test.");
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -60,6 +60,23 @@ fn run() -> Result<()> {
                 .help("An alias for -n/--length"),
         )
         .arg(
+            Arg::with_name("range")
+                .short("r")
+                .long("range")
+                .takes_value(true)
+                .value_name("N:M")
+                .conflicts_with("length")
+                .conflicts_with("bytes")
+                .help("Read only bytes N through M from the input")
+                .long_help(
+                    "Only print the specified range of bytes. \
+                     For example:\n  \
+                     '--range 512:1024' prints bytes 512 to 1024\n  \
+                     '--range 512:+512' skips 512 bytes and prints the next 512 bytes (equivalent to 512:1024)\n  \
+                     '--range :512' prints the first 512 bytes\n  \
+                     '--range 512:' skips 512 bytes and prints the rest of the input")
+        )
+        .arg(
             Arg::with_name("nosqueezing")
                 .short("v")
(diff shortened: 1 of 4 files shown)


sharkdp:

From a first glance, the changes look great, but I didn't have the time for a full review yet. Could be a few more days.

src/bin/hexyl.rs
```diff
@@ -94,6 +129,19 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
...
     }
 
+    let byterange_arg = matches.value_of("range");
+    let mut range_offset = 0;
+    if let Some(range) = byterange_arg {
+        if let Ok((offset, num_bytes)) = parse_range(range) {
+            range_offset = offset;
+            let mut discard = vec![0u8; offset as usize];
+            reader
+                .read_exact(&mut discard)
```
sharkdp: Hm, that seems a bit hacky, but I also couldn't find a better solution after a quick search.

I'm not sure if we should leave it like this. If somebody wants to skip the first few MB or even GB of a file/input, we would actually allocate a buffer of this size on the heap which could easily fail. Even if it doesn't fail, it seems wrong to allocate a huge buffer for simply discarding everything inside it.

If that helps, I think we could actually use the `BufRead` trait instead of just `Read`. We would simply need to create a `BufReader` for the input file. The `StdinLock` already implements `BufRead`.

sharkdp: What we want is `Seek`: https://doc.rust-lang.org/stable/std/io/struct.BufReader.html#impl-Seek

src/bin/hexyl.rs
```diff
@@ -140,27 +190,108 @@ fn main() {
...
+        assert_eq!(offset, 512);
+        assert_eq!(bytes_to_read, 0);
+
+        let (offset, bytes_to_read) = parse_range("512:+512").expect("Not allowed to fail test.");
+        assert_eq!(offset, 512);
+        assert_eq!(bytes_to_read, 512);
+    }
+
+    #[test]
+    fn parse_bad_input() {
```
sharkdp: What happens if the range arg is completely empty?
```
--range ""
```

src/bin/hexyl.rs
```diff
@@ -140,27 +190,108 @@ fn main() {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_empty_range() {
+        let (offset, bytes_to_read) = parse_range(":").expect("Not allowed to fail test.");
+        assert_eq!(offset, u64::min_value());
```
sharkdp: I would probably replace `u64::min_value()` by `0` or `0u64` (in all tests).

src/bin/hexyl.rs
MaxJohansen: This is the same as providing an empty range:
```
--range ":"
```

I'll add it to the test :)

src/bin/hexyl.rs
MaxJohansen: I did a bit of research on this and Stdin doesn't appear to be seekable no matter what I do. I'll continue exploring other options for this.

bjorn3: Maybe read to a 1024 bytes or so big slice multiple times as necessary and then read to a smaller slice as necessary.

sharkdp:
I'm going to close this for now, with #88 merged. Happy to reopen it if there is any further comments/progress.

PR #76, opened 2019-10-24
Scarf install instructions

aviaviavi:
Hello! This change adds install instructions with the [Scarf](https://scarf.sh) package manager (disclaimer: I'm the author of Scarf).

I saw that hexyl is being supported on Github Sponsors. Scarf could be an additional way to support this project, especially from those using it in a commercial setting. It will collect and provide you with usage statistics as well as offer the ability to collect payments to opt out, so anyone who installs hexyl in this way will be supporting the project in one form or another.

I'd be happy to transfer package ownership and/or co-maintain it, so no extra work required on your part.

I also opened a similar PR on `fd`, so apologies for the noise if you already saw that 😄


sharkdp:
Thank you for your offer. I am currently not planning to monetarize my open source projects in this way.

PR #86, opened 2020-03-14
snap support

purvesh-linux11:
A **candidate/stable release snap** is created for cross platform installation among various Linux distros, including CentOS, KDE Neon, openSUSE, Red Hat and many more.

snap is already live on : https://snapcraft.io/hexyl

you can check out that.


bjorn3:
You should probably not commit the snap, but instead setup CI to build it.

sharkdp:

As @bjorn3 stated, we don't really want to add a binary snap file to this repository. Also, as I am not maintaining the `snap` file myself (and not planning to), I would prefer if distribution-specific files would be hosted somewhere else (e.g. in a separate repository).

I didn't really understand why you would like to add additional `testdata` files?


What we can definitely do is to add installation instructions for `snap` to the README.

purvesh-linux11:
@sharkdp , thanks for check out my pull request. 
I understand your concern and i am very happy to host a separate repository for snap distribution of this repo.

sharkdp:
@purvesh-linux11 Thank you!

I'm going to close this PR. Happy to include the README changes at some point.

Issue #87, opened 2020-04-29
'skip' option


Tarnadas:
I'd like to have an option to skip n bytes.

Examples:

`hexyl --skip 0x20`
Prints bytes from pos 0 to 0x20

`hexyl --skip 0x50 --length 0x100`
Prints bytes from pos 0x100 to 0x150

Tarnadas:
sorry, it is obviously a duplicate of #16

PR #88, opened 2020-04-29
feat: add skip option

### LINKED ISSUE #16: Options for reading between two offsets

Tarnadas:
Since PR #75 has been open for quite some time now, I wanted to implement my own `skip` option.
This PR uses the `Seek` trait as suggested.

resolves #16 and #87

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -18,10 +18,21 @@ impl<'a> Read for Input<'a> {
 impl<'a> Seek for Input<'a> {
     fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
         match *self {
-            Input::File(ref mut file) => file.seek(pos),
+            Input::File(ref mut file) => {
+                let seek_res = file.seek(pos);
+                if let Err(Some(29)) = seek_res.as_ref().map_err(|err| err.raw_os_error()) {
+                    return Err(io::Error::new(
+                        io::ErrorKind::Other,
+                        "Using '--seek' is not supported when using a pipe",
+                    )
+                    .into());
+                };
+                dbg!(&seek_res);
+                seek_res
+            }
             Input::Stdin(_) => Err(io::Error::new(
                 io::ErrorKind::Other,
-                "not supported by stdin-input",
+                "Using '--seek' is not supported when reading from STDIN",
             )),
         }
     }
--- a/src/input.rs
+++ b/src/input.rs
@@ -20,14 +20,14 @@ impl<'a> Seek for Input<'a> {
         match *self {
             Input::File(ref mut file) => {
                 let seek_res = file.seek(pos);
-                if let Err(Some(29)) = seek_res.as_ref().map_err(|err| err.raw_os_error()) {
+                if let Err(Some(libc::ESPIPE)) = seek_res.as_ref().map_err(|err| err.raw_os_error())
+                {
                     return Err(io::Error::new(
                         io::ErrorKind::Other,
                         "Using '--seek' is not supported when using a pipe",
                     )
                     .into());
                 };
-                dbg!(&seek_res);
                 seek_res
             }
             Input::Stdin(_) => Err(io::Error::new(
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,17 +1,16 @@
-use std::cell::RefCell;
 use std::fs;
 use std::io::{self, Read, Seek, SeekFrom};
 
 pub enum Input<'a> {
     File(fs::File),
-    Stdin(RefCell<io::StdinLock<'a>>),
+    Stdin(io::StdinLock<'a>),
 }
 
 impl<'a> Read for Input<'a> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         match *self {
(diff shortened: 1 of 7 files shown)


Tarnadas:
@sharkdp Should I try to make it compile on Rust 1.31 or would you rather like to increase minimum required Rust version?

sharkdp:
Thank you for your contribution. I bumped the min. required Rust version to 1.36 and "backported" it to this version (by matching on `Input::*` instead of `Self::*`).

I also removed the `RefCell` layer as it is not needed.

src/input.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+impl<'a> Seek for Input<'a> {
+    fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
+        match *self {
+            Input::File(ref mut file) => file.seek(pos),
+            Input::Stdin(_) => Err(io::Error::new(
+                io::ErrorKind::Other,
+                "not supported by stdin-input",
```
sharkdp, line 24: I like the general idea, but the error message could be slightly improved. Maybe:
*"Using '--seek' is not supported when reading from STDIN."*

sharkdp:
This will also fail if the input is not a file, but a pipe, for example (this is not "reading from STDIN"):
```
❯ hexyl <(echo test) --skip 2
Error: Illegal seek (os error 29)
```
Maybe we could catch that specific OS error and print a similar error message.


Another thing: when using `--skip N`, shouldn't the display offset be set to `N`? (unless the user explicitly specifies it).

Tarnadas:
I changed the error message to what you suggested. It will now also catch this specific OS error and display a proper error message.

You're right about the display offset, so it will now use skip as a default, if the user does not specifiy it.

It should still be possible to implement the `--range N:M` option. I think it would even make it easier, because you basically just have to parse it and set skip and length respectively.

src/input.rs
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+}
+
+impl<'a> Seek for Input<'a> {
+    fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
+        match *self {
+            Input::File(ref mut file) => {
+                let seek_res = file.seek(pos);
+                if let Err(Some(29)) = seek_res.as_ref().map_err(|err| err.raw_os_error()) {
```
sharkdp, line 23: can we use [`libc::ESPIPE`](https://docs.rs/libc/0.2.70/libc/constant.ESPIPE.html) here instead of 29?

src/input.rs
```diff
@@ -0,0 +1,48 @@
...
+                    return Err(io::Error::new(
+                        io::ErrorKind::Other,
+                        "Using '--seek' is not supported when using a pipe",
+                    )
+                    .into());
+                };
+                dbg!(&seek_res);
```
sharkdp, line 30: Leftover from debugging?

Tarnadas: removed it

src/input.rs
Tarnadas: yeah, good idea

PR #90, opened 2020-05-20
Add unit system (closes #44)

### LINKED ISSUE #44: Units of measurement of bytes?
| Unit    | Implemented?      | Description                                                                                                                                                                     | Examples                                                                             | Suggested implementation                                                                                                    |
| ------- | ------------      | -------------------------------------------------------------------------------                                                                                                 | --------------------------------------------------------------------------------     | -----------------------------------------------------------------------------------------------------------------------     |
| Decimal | [x]               | A decimal integer, which is equivalent to specifying a single byte unit for the count.                                                                                          | `23`, `1024`                                                                         | `u64::from_str(...)`                                                                                                        |
| Hex     | [x] Implemented in #45. | A hexadecimal integer. Specified with a leading `0x`.                                                                                                                           | `0x17`, `0x100`                                                                      | `u64::from_str_radix(...)`                                                                                                  |
| Blocks  | [ ]               | A single block, which is by default 512 bytes but configurable via config flag.                                                                                                 | `-b 512 -n 1block`<br><br>N.B: one cannot use a block unit to define the block size. | Add a flag to optionally define block size, then check for a trailing `block` when parsing numbers. Multiply by block size. |
| Bytes   | [ ]               | A byte size familiar to most IT professionals. Specified by `B` at the end of the count, and can include an optional magnitudinal spec like kilobytes (`K`) or megabytes (`M`). | <ul><li>`23B`: 23 bytes</li><li>`9KB`: 9 kilobytes</li></ul>                         | Implement a regex of the form `(?P<count>\d+)(?P<magnitude_unit>[KM]?)B`.                                                   |

## Other open questions

* **Should the block unit, which might be controversial, be scoped into another issue?**
* **Should a single leading `-` or `+` sign be supported?**
	* Not sure how `+` is useful -- `xxd`'s manual states that for [the `-s` option](https://linux.die.net/man/1/xxd) `+` is useful only for `stdin`. Not sure what that means, though.

ErichDonGubler:
This PR is conceptually similar to @aswild's commits mentioned [here](https://github.com/sharkdp/hexyl/issues/44#issuecomment-631085679), except we attempt to parse units *at the first non-ASCII digit*, rather than starting from the first ASCII alphabetical. Tests have been shamelessly stolen from those commits.

There are several other housekeeping commits I've added here. If desired, I can split these out into a separate PR.

### DIFF
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -1,8 +1,6 @@
 #[macro_use]
 extern crate clap;
 
-use atty;
-
 use std::fs::File;
 use std::io::{self, prelude::*, SeekFrom};
 
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -163,7 +163,11 @@ fn main() {
 
 fn parse_hex_or_int(n: &str) -> Option<u64> {
     if n.starts_with("0x") {
-        u64::from_str_radix(n.trim_start_matches("0x"), 16).ok()
+        let n = n.trim_start_matches("0x");
+        if n.chars().next() == Some('+') {
+            return None;
+        }
+        u64::from_str_radix(n, 16).ok()
     } else {
         n.parse::<u64>().ok()
     }
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -44,6 +44,16 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
                 .help("Skip first N bytes"),
         )
         .arg(
+            Arg::with_name("block_size")
+                .long("block-size")
+                .takes_value(true)
+                .value_name("SIZE")
+                .help(
+                    "Sets the size of the `block` unit to SIZE. Examples: \
+                    --block-size=1024, --block-size=4kB",
+                ),
+        )
+        .arg(
             Arg::with_name("nosqueezing")
                 .short("v")
                 .long("no-squeezing")
@@ -91,7 +101,14 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
         None => Input::Stdin(stdin.lock()),
     };
 
-    let skip_arg = matches.value_of("skip").and_then(parse_byte_count);
+    let block_size = matches
+        .value_of("block_size")
+        .and_then(|bs| bs.parse::<u64>().ok())
+        .unwrap_or(512);
+
+    let skip_arg = matches
+        .value_of("skip")
+        .and_then(|s| parse_byte_count(s, block_size));
 
     if let Some(skip) = skip_arg {
         reader.seek(SeekFrom::Start(skip))?;
@@ -101,7 +118,8 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
         .value_of("length")
         .or_else(|| matches.value_of("bytes"));
 
-    let mut reader = if let Some(length) = length_arg.and_then(parse_byte_count) {
+    let mut reader = if let Some(length) = length_arg.and_then(|s| parse_byte_count(s, block_size))
+    {
         Box::new(reader.take(length))
     } else {
         reader.into_inner()
@@ -123,7 +141,7 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
 
     let display_offset = matches
         .value_of("display_offset")
-        .and_then(parse_byte_count)
+        .and_then(|s| parse_byte_count(s, block_size))
         .unwrap_or_else(|| skip_arg.unwrap_or(0));
 
     let stdout = io::stdout();
@@ -161,7 +179,7 @@ fn main() {
     }
 }
 
-fn parse_byte_count(n: &str) -> Option<u64> {
+fn parse_byte_count(n: &str, block_size: u64) -> Option<u64> {
     const HEX_PREFIX: &'static str = "0x";
 
     let n = {
@@ -196,6 +214,7 @@ fn parse_byte_count(n: &str) -> Option<u64> {
                     ("mib", 1024u64.pow(2)),
                     ("gib", 1024u64.pow(3)),
                     ("tib", 1024u64.pow(4)),
+                    ("block", block_size),
                 ]
                 .iter()
                 .cloned()
@@ -218,13 +237,16 @@ fn parse_byte_count(n: &str) -> Option<u64> {
 fn test_parse_byte_count() {
     macro_rules! success {
         ($input: expr, $expected: expr) => {
-            assert_eq!(parse_byte_count($input), Some($expected));
+            success!($input, 512, $expected)
+        };
+        ($input: expr, $block_size: expr, $expected: expr) => {
+            assert_eq!(parse_byte_count($input, $block_size), Some($expected));
         };
     }
 
     macro_rules! error {
         ($input: expr) => {
-            assert_eq!(parse_byte_count($input), None);
+            assert_eq!(parse_byte_count($input, 512), None);
         };
     }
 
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you for your contribution!

See comments in #44

> Tests have been shamelessly stolen from those commits.

@aswild: I am assuming that you are okay with this? :smile:

ErichDonGubler:
I would be more than happy to make @aswild a co-author for the unit system commits -- I've only just learned how to do it on Git, which is why I didn't do so already.

bjorn3:
The following my help: https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

ErichDonGubler:
> The following my help: https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

Yep, that's what I just learned how to do! :)

ErichDonGubler:
I've pushed `FIXUP` changes removing `{k,m,g,t}` units and the `-b` alias for `--block-size` that ~~can~~ should be squashed once the review is complete and @sharkdp wants to merge this -- CC #44.

ErichDonGubler:
I have implemented the reviewed feedback with `FIXUP` commits that should be squashed before merge. What say you, @sharkdp?

aswild:
> @aswild: I am assuming that you are okay with this? smile

Yeah I'm cool with that, and don't even need explicit author/contributor credit - the tests were just a basic list of good/bad values I thought of off the top of my head.

src/bin/hexyl.rs
```diff
@@ -93,15 +100,24 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
...
+
+    let skip_arg = matches
+        .value_of("skip")
+        .and_then(|s| parse_byte_count(s, block_size));
 
     if let Some(skip) = skip_arg {
         reader.seek(SeekFrom::Start(skip))?;
     }
 
+    #[allow(clippy::or_fun_call)]
```
aswild, line 116: Why disable the clippy lint rather than just fixing it with `.or_else(|| matches.value_of("bytes"))`? (same below)

src/bin/hexyl.rs
```diff
@@ -161,10 +178,127 @@ fn main() {
...
+        let mut chars = n.chars();
+        match chars.next()? {
+            '+' => chars.as_str(),
+            _ => n,
+        }
+    };
+
+    if n.starts_with(HEX_PREFIX) {
+        let n = &n[HEX_PREFIX.len()..];
+        if !n.chars().all(|c| match c {
```
aswild, line 194: this validation can be skipped, `u64::from_str_radix` will do it anyway

src/bin/hexyl.rs
ErichDonGubler: I simply disabled the lint as a conservative effort to preserve the eagerness of the expression being evaluated. I'm not positive a behavior change would result, but I just avoided the question altogether.

src/bin/hexyl.rs
ErichDonGubler: From the commit message:

> This is different from simply using `from_str_radix` because `from_str_radix` accepts signs (in this case, only `'+'`).

ErichDonGubler: Thinking about this some more, it would probably be just as well to check for a `'+'` as the first character, and let `u64::from_str_radix` fail the other cases. That would better match the intent of this code.

ErichDonGubler: Changed. What do you think now? :)

src/bin/hexyl.rs
sharkdp: Using `.or_else` is fine here.

src/bin/hexyl.rs
```diff
@@ -45,6 +45,13 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
                 .value_name("N")
                 .help("Skip first N bytes"),
         )
+        .arg(
+            Arg::with_name("block_size")
+                .long("block-size")
+                .takes_value(true)
+                .value_name("N")
+                .help("Sets `block` unit to N"),
```
sharkdp, line 53: minor: *"… unit to N bytes"*?.

src/bin/hexyl.rs
```diff
@@ -121,9 +137,10 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
 
     let squeeze = !matches.is_present("nosqueezing");
 
+    #[allow(clippy::or_fun_call)]
```
sharkdp, line 140: Same here. I'm not a fan of these `allow(clippy::…)` lines. Either we ignore clippy warnings that are completely irrelevant, or fix them.

src/bin/hexyl.rs
```diff
@@ -161,10 +178,124 @@ fn main() {
...
+        if n.chars().next() == Some('+') {
+            return None;
+        }
+        return u64::from_str_radix(n, 16).ok();
+    }
+
+    let (n, unit_multiplier) = match n.chars().position(|c| !c.is_ascii_digit()) {
+        Some(unit_begin_idx) => {
+            let (n, raw_unit) = n.split_at(unit_begin_idx);
+            let raw_unit = Ascii::new(raw_unit);
```
sharkdp, line 203: do we need a new dependency for this? Couldn't we just use `unit == raw_unit.to_lower()`?

ErichDonGubler: I'm used to approaching performance first, which was why I reached for `unicase` out of habit, but the fact is nobody will likely notice an allocation to get a lowercase string here. If you'd prefer that over another dep, sounds valid to me!

src/bin/hexyl.rs
ErichDonGubler: Sure. I can't see any reason why there would be a regression, but I figured letting this get caught in review was better than me overthinking it. :)

src/bin/hexyl.rs
ErichDonGubler: Good catch. I'll fix this.

Actually, this raises an interesting question to me.  It could be useful to interpret a byte offset here instead of just straight bytes, since often block sizes are multiples of `kb`. That can come later, if necessary, but it seems just as easy to do now. What do you think of applying `parse_byte_offset` here?

ErichDonGubler:
> Yeah I'm cool with that, and don't even need explicit author/contributor credit - the tests were just a basic list of good/bad values I thought of off the top of my head.

You did a significant amount of effort, not even taking into account the code you wrote in your own implementation. Your efforts were important, so I'm gonna make you at least a co-author for the commits. 😉

src/bin/hexyl.rs
sharkdp: I would prefer the easier solution without the extra crate. This will also make it easier for package maintainers (for distributions that package every dependency, like Fedora or Debian).

src/bin/hexyl.rs
sharkdp: That makes total sense, yes :+1: In that case, the help text can either stay this way or maybe rename `N` to `SIZE` and then:

*"Set the size of the \`block\` unit to SIZE. Examples: --block-size=1024, --block-size=4kB"*

ErichDonGubler: Done. The only modification I made to your suggestion is to use indicative tense ("Sets") instead of imperative ("Set"), since that's the tense most of the options use.

src/bin/hexyl.rs
ErichDonGubler: Fixed.

src/bin/hexyl.rs
ErichDonGubler: Fixed.

src/bin/hexyl.rs
ErichDonGubler: I've removed `unicase` and now use `str::to_lowercase` instead.

ErichDonGubler:
@aswild @sharkdp I've resolved the issues that I feel have a conclusive end, just one left for each of you for concerns you've brought up.

PR #94, opened 2020-06-01
BREAKING CHANGE(lib): Use `u64` for `Printer::display_offset` and `Squeezer::process`, not `usize`

ErichDonGubler:
~~Soft dependency on #93.~~

Technically a breaking change for architectures where `size_of::<usize>() > size_of::<u64>()`, but that strikes me as extremely unlikely.

This change is important because most file seek offsets, even on older architectures with word sizes smaller than `u32`, accept a `u64`. In particular, Rust's own `std` acknowledges this. So, we're actually going against prior art in Rust by using a `usize`.


sharkdp:
Closing, as all changes have been merged with #95

PR #95, opened 2020-06-01
Take advantage of skip behavior from #93 for `--seek` for non-`Seek` `Input`s

ErichDonGubler:
~~Hard dependency on #93~~, less hard on #94.

Prior to this commit, `Input::seek` would fail under the following conditions:

1. For most reasonable I/O failure conditions, like a file being moved or truncated, or the filesystem being interrupted, or...*voodoo*. /me waves hands in the air
2. When attempting to call `File::seek` on a file backed by a pipe (i.e., a FIFO or socket), as noted by the `ESPIPE` error check.
3. When called on a `Stdin` variant, because `StdinLock` doesn't implement `Seek` -- conceptually similar to case 2.

With the lib commit, 2 and 3 are changed to simulate a `SeekFrom::Current(positive_value)` by throwing away `positive_value` bytes by `std::io::copy`ing into `std::io::sink`. With the commit changing the binary, we use `SeekFrom::Current` (instead of `SeekFrom::Start`), since the "current" offset will always be 0 anyway. Speaking in terms of the command line, doing something like this failed before:

```sh
echo 'asdf' | hexyl --skip 2 # hoping to just print 'df'
# Errors out, because we can't `seek` on `Input::Stdin`
```

...but now it should work as intended.

---

This is technically a breaking change because it halves the upper range of accepted values for byte counts. I thought that fine under the assumption it will be rare for users to ever want to specify something in the upper half of `u64`'s value range, especially considering that this can pave the way for having negative byte offsets (instead of just counts).

### DIFF
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -1,6 +1,7 @@
 #[macro_use]
 extern crate clap;
 
+use std::convert::TryFrom;
 use std::fs::File;
 use std::io::{self, prelude::*, SeekFrom};
 
@@ -103,15 +104,15 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
 
     let block_size = matches
         .value_of("block_size")
-        .and_then(|bs| bs.parse::<u64>().ok())
-        .unwrap_or(512);
+        .and_then(|bs| bs.parse().ok().and_then(PositiveI64::new))
+        .unwrap_or_else(|| PositiveI64::new(512).unwrap());
 
     let skip_arg = matches
         .value_of("skip")
         .and_then(|s| parse_byte_count(s, block_size));
 
-    if let Some(skip) = skip_arg {
-        reader.seek(SeekFrom::Start(skip))?;
+    if let Some(skip) = skip_arg.clone() {
+        reader.seek(SeekFrom::Current(skip.into_inner()))?;
     }
 
     let length_arg = matches
@@ -120,7 +121,7 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
 
     let mut reader = if let Some(length) = length_arg.and_then(|s| parse_byte_count(s, block_size))
     {
-        Box::new(reader.take(length))
+        Box::new(reader.take(length.into()))
     } else {
         reader.into_inner()
     };
@@ -143,7 +144,8 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
         .value_of("display_offset")
         .and_then(|s| parse_byte_count(s, block_size))
         .or(skip_arg)
-        .unwrap_or(0);
+        .unwrap_or_default()
+        .into();
 
     let stdout = io::stdout();
     let mut stdout_lock = stdout.lock();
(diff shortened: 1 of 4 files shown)


src/bin/hexyl.rs
```diff
@@ -179,23 +182,50 @@ fn main() {
...
+        if x.is_negative() {
+            None
+        } else {
+            Some(Self(x))
+        }
+    }
+
+    pub fn into_inner(self) -> i64 {
+        let Self(x) = self;
+        x
```
sharkdp, line 199: This could also be written as `self.0`.

src/bin/hexyl.rs
```diff
@@ -179,23 +182,50 @@ fn main() {
...
+    }
+}
+
+impl Into<u64> for PositiveI64 {
+    fn into(self) -> u64 {
+        let Self(x) = self;
+        u64::try_from(x).unwrap()
```
sharkdp, line 206: Minor: I like to use `.expect(…)` instead of `.unwrap()`, if only for documentation purposes. Something like:
```rust
try_from(x).expect("Inner value is guaranteed to be positive")
```

src/bin/hexyl.rs
```diff
@@ -179,23 +182,50 @@ fn main() {
     }
 }
 
-fn parse_byte_count(n: &str, block_size: u64) -> Option<u64> {
+#[derive(Clone, Copy, Debug, Default, Hash, Eq, Ord, PartialEq, PartialOrd)]
+pub struct PositiveI64(i64);
```
sharkdp, line 186: :+1:

src/input.rs
```diff
@@ -17,23 +18,39 @@ impl<'a> Read for Input<'a> {
 
 impl<'a> Seek for Input<'a> {
     fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
+        fn try_skip<R>(reader: R, pos: SeekFrom, err_desc: &'static str) -> io::Result<u64>
```
sharkdp, line 21: Why is this implemented as a nested function instead of a separate function in the `impl` block? If you want to keep it as a inline function, maybe capture `pos` instead of passing it in twice?

ErichDonGubler: I would have made this a closure, except closures don't allow generics. Boo. I kept it as a nested function to keep the scope conservative -- but there's no issue with moving it to the `impl` scope, if you want.

src/bin/hexyl.rs
ErichDonGubler: :+1:

src/bin/hexyl.rs
ErichDonGubler: It could. In general, I like to destructure (to make refactoring easier), but I DO prefer codebase consistency and didn't take the time to examine what other chunks of the codebase are doing, oops. I can change that here, since it's very unlikely things will change in the future.

src/input.rs
sharkdp: Ah, I see - thanks. Let's keep it as is.

ErichDonGubler:
> I don't quite understand how the "only support seeking forward with a relative offset" would be triggered?

Uh, do you mean how we'd exercise the success path, or the error message with similar contents? The explanation I have will answer for both, so...I'll just get on with it:

Prior to this commit, `Input::seek` would fail under the following conditions:

1. For most reasonable I/O failure conditions, like a file being moved or truncated, or the filesystem being interrupted, or...*voodoo*. /me waves hands in the air
2. When attempting to call `File::seek` on a file backed by a pipe (i.e., a FIFO or socket), as noted by the `ESPIPE` error check.
3. When called on a `Stdin` variant, because `StdinLock` doesn't implement `Seek` -- conceptually similar to case 2.

With the lib commit, 2 and 3 are changed to simulate a `SeekFrom::Current(positive_value)` by throwing away `positive_value` bytes. With the commit changing the binary, we use `SeekFrom::Current` (instead of `SeekFrom::Start`), since the "current" offset will always be 0 anyway. Speaking in terms of the command line, doing something like this failed before:

```sh
echo 'asdf' | hexyl --skip 2 # hoping to just print 'df'
# Errors out, because we can't `seek` on `Input::Stdin`
```

...but now it should work as intended.

ErichDonGubler:
When writing the error message, I tried to write the most concise and correct explanation...but that's a terrible heuristic for diagnostics. If there's something better, let's bend the curve! I'm not so sure I'm the best to come up with a descriptive error message, though, since I've gotten so close to the problem. I'll think about alternatives.

src/bin/hexyl.rs
ErichDonGubler: Updated -- how's it look?

src/bin/hexyl.rs
ErichDonGubler: I've updated to use `self.0`.

sharkdp: > In general, I like to destructure (to make refactoring easier),

I see. In this case I would probably just go with a regular struct instead of a tuple and then return `self.inner` or similar. But thank you for the explanation!

> but I DO prefer codebase consistency and didn't take the time to examine what other chunks of the codebase are doing, oops

To be honest: I'm not particular proud of `hexyl`s code base. I have never had to time to clean things up before (and after) contributions starting coming in. In this sense, there's not really one style of coding being used here. But thank you for considering this aspect!

sharkdp:
> Speaking in terms of the command line, doing something like this failed before:
> 
> ```shell
> echo 'asdf' | hexyl --skip 2 # hoping to just print 'df'
> # Errors out, because we can't `seek` on `Input::Stdin`
> ```
> 
> ...but now it should work as intended.

This is the part I understood. But how could the "only support seeking forward with a relative offset" be triggered by a user of `hexyl`?

PR #124, opened 2020-11-19
Add installation instructions for Guix

foo-dogsquared:
I've added Hexyl to the list of packages for Guix (https://guix.gnu.org/en/packages/hexyl-0.8.0/) so I thought to add it here to show GNU has a similar NixOS-inspired system.


sharkdp:
Thank you!

PR #133, opened 2021-04-23
Fix #131

### LINKED ISSUE #131: --version prints newline to stderr
This is a bit weird, not sure if intentional:

```
~ $ hexyl --version
hexyl 0.8.0
~ $ hexyl --version > /dev/null

~ $ hexyl --version &> /dev/null
~ $
```

scimas:
I noticed that the `eprint` in `main` itself is taking care of the `--help` output, but the `--version` output is coming without the `eprint`. So I have done two changes:

-Don't print the error for version output, instead add a newline print.
-Move the eprint for help only.


sharkdp (approved):

PR #144, opened 2021-11-07
Allow hex numbers and data units for block size argument

### LINKED ISSUE #111: Argument `--block-size` does not accept units (or hex numbers), contrary to its helptext description.
The `hexyl --help` description for the `--block-size` argument goes as follows:

```
        --block-size <SIZE>     Sets the size of the `block` unit to SIZE.
                                Examples: --block-size=1024, --block-size=4kB
```

However, when I attempt to run `hexyl "$(which hexyl)" -n 1block --block-size=4kB`, hexyl prints the following error:

```
Error: failed to parse `--block-size` arg "4kb" as positive integer

Caused by:
    invalid digit found in string
```

This is caused by the `block_size` argument being parsed as a positive i64 number, rather than as a data unit:

https://github.com/sharkdp/hexyl/blob/cc5b308fc9c2ca57ba176cf69a237f13428bb8e3/src/bin/hexyl.rs#L133-L142

**Side note:** If this bug was fixed, how would hexyl handle `--block-size=1block`? Should it be multiplying it by the default value, or would it be more appropriate to treat as an error to avoid any confusion?

**Edit:** It also doesn't accept hexadecimal numbers.

---

Note: I'm doing a class assignment regarding creating tests (from scratch) for an open-source command line program, and I chose hexyl for it. You might have a couple more of these on the way, depending on what the assignment brings to light :)

merkrafter:
## Summary
Closes #111 
This PR allows arguments like `4kb` or `0x10` for `--block-size`.
The 'block' unit is not allowed in this context (see discussion on related issue).


sharkdp:

This looks great on a first glance. Happy to review it in detail after the cleanup.

> I used the [`const_format`](https://lib.rs/crates/const_format) crate to display the default block size to the user in a convenient and consistent way. However, using it required Rust 1.46.0+. Do you think this detail is worth raising the minimum required Rust version?

I didn't know this crate. That sounds really helpful. Bumping to Rust 1.46 sounds good. Since this only affects hexyl as a tool and not hexyl as a library, I'd be okay to bump it even further, in case that helps with other parts of the code. I don't think we have a lot of "customers" anyway: https://crates.io/crates/hexyl/reverse_dependencies

merkrafter:
Please let me know if you have other concerns/ideas/suggestions etc. :))

sharkdp (approved):
Thank you very much. Great work on the refactoring - this looks much cleaner now! I looked closely but I don't have any additional comments ;-)

PR #119, opened 2020-10-30
bin: offset should add to skip, not override it

tommilligan:
Closes #115

This PR aligns the behaviour of `hexyl` with `xxd`, when the `--skip` and `--display-offset` flags are both set.

Currently, `--display-offset` overrides `--skip` . In `xxd`, the two values are added together to produce the first address value.

[Comparison of current `hexyl` and `xxd` behaviour.](https://github.com/sharkdp/hexyl/issues/115#issuecomment-719823848)

PR #145, opened 2021-11-16
Allow negative values for `--display-offset`

### LINKED ISSUE #112: Argument `--display-offset` does not accept negative numbers, contrary to its helptext description.
The hexyl --help description for the --block-size argument goes as follows:

```
    -o, --display-offset <N>    Add N bytes to the displayed file position. The
                                N argument can also include a unit (see
                                `--length` for details)
                                A negative value is valid and calculates an
                                offset relative to the end of the file.
```

However, when I attempt to run `dd if=/dev/urandom bs=16 count=32 &>/dev/null | hexyl --skip=0xFF --display-offset=-0x2F`, hexyl prints the following error:

```
Error: failed to parse `--display-offset` arg "-0x2F" as byte count

Caused by:
    negative offset specified, but only positive offsets (counts) are accepted in this context
```

---

Note: I'm doing a class assignment regarding creating tests (from scratch) for an open-source command line program, and I chose hexyl for it. You might have a couple more of these on the way, depending on what the assignment brings to light :)

ErichDonGubler:
`--display-offset` has had an odd relationship with negated inputs over its lifetime. Historical behaviors (and the addition being proposed here) might be easiest to understand in a tabular format.

| When | Behavior summary | Negative offset behavior |
| ---- | --------| --- | 
| First implemented in #72. | Absolute position, positive values only. | Disallowed. |
| Refined to allow negatives in #99. | Absolute position, including then-new semantics for negatives. | Allowed, end offset backwards from EOF (and so only usable on files), prompting #112 for poor diagnostics at the time. |
| Actual compatibility with `xxd` desired, made an offset of `--skip` in #119. | Relative offset based on `--skip`. | Disallowed. |

**This PR proposes to make negative values for `--display-offset` meaningful again by allowing them again when `--skip` is specified, adding a negative value to `--skip` if provided.** To use the same format as above:

| When | Behavior summary | Negative offset behavior |
| ---- | --------| --- | 
| **This PR.** | Relative offset based on `--skip`. | **Allowed when `--skip` is specified; underflow results in an error.** |

More detailed overview of the design and rationale: 

* This stays consistent with current behavior -- the final display offset retains its concept as "`--skip` + `--display-offset`".
* This adds an error case for the display offset underflowing when added to the unsigned `--skip` offset:

    ```
    hexyl --display-offset=-21 --skip 20 .\some_file
    Error: invalid value when applying `--display-offset` to `--skip`

    Caused by:
        base (20) - offset (21) < 0
    ```
* Bonus: the error case where offset addition overflows now has a nice diagnostic:

    ```
    TODO
    ```

* If `--skip` is not specified, it is an error, as before:

    ```
    Error: failed to parse `--display-offset` arg "-1" as byte count

    Caused by:
        negative offset specified, but only positive offsets (counts) are accepted in this context
    ```

Resolves #112.

---

PR checklist:

- [ ] Get deets from PR OP into commit messages for history's sake.
- [ ] Tests

### DIFF
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -2,6 +2,7 @@
 extern crate clap;
 
 use std::convert::TryFrom;
+use std::fmt::{self, Display, Formatter};
 use std::fs::File;
 use std::io::{self, prelude::*, SeekFrom};
 use std::num::NonZeroI64;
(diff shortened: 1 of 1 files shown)


sharkdp:
Sounds great. Thank you very much for looking into this.

sharkdp:
I'm closing this due to inactivity. Please feel free to comment in case it should be re-opened.

PR #151, opened 2022-03-22
Add man page

sorairolake:
This is written in ~~AsciiDoc or~~ Pandoc’s Markdown.

These can be converted by:

- ~~`asciidoctor -b manpage doc/hexyl.1.adoc`~~
- `pandoc -s -f markdown -t man -o doc/hexyl.1 doc/hexyl.1.md`

I will fix to add a conversion process to the CI, add a troff file as well, keep only it, or other. Which one is better?

### DIFF
--- a/doc/hexyl.1.md
+++ b/doc/hexyl.1.md
@@ -1,0 +1,145 @@
+% HEXYL(1) hexyl 0.9.0 | General Commands Manual
+%
+% 2022-03-23
+
+<!--
+This man page is written in Pandoc's Markdown.
+See: https://pandoc.org/MANUAL.html#pandocs-markdown
+-->
+
+# NAME
+
+hexyl - a command-line hex viewer
+
+# SYNOPSIS
+
+**hexyl** [_OPTIONS_] [_FILE_]
+
+# DESCRIPTION
+
+**hexyl** is a simple hex viewer for the terminal.
+It uses a colored output to distinguish different categories of bytes (NULL
+bytes, printable ASCII characters, ASCII whitespace characters, other ASCII
+characters and non-ASCII).
+
+# POSITIONAL ARGUMENTS
+
+_FILE_
+:   The file to display.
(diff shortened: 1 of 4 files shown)


sharkdp:
Awesome. Thank you very much for working on this. I would prefer the pandoc Markdown => man page way because I am familiar with this from other projects. And I don't know AsciiDoc.

sorairolake:
@sharkdp OK. Added the conversion process from Pandoc's Markdown in the CI. Also, I wrote the conversion method in the README.

doc/hexyl.1.md
```diff
@@ -0,0 +1,145 @@
...
+
+    Examples:
+
+    :   
+
+        Sets the block size to 1024 bytes:
+        :   $ **hexyl \--block-size=1024**
+
+        Sets the block size to 4 kilobytes:
+        :   $ **hexyl \--block-size=4kB**
```
sharkdp, line 73: Setting block size alone is not really that interesting. It's only relevant if you use it in one of the other command-line options. As in:
```
hexyl --block-size=1024 --length=5block …
```

sharkdp (commented):
Thank you! I added one minor comment.

doc/hexyl.1.md
sorairolake: Fixed.

doc/hexyl.1.md
```diff
@@ -0,0 +1,145 @@
...
+
+    Examples:
+
+    :   
+
+        Sets the block size to 1024 bytes:
+        :   $ **hexyl \--block-size=1024 \--length=5block**
+
+        Sets the block size to 4 kilobytes:
+        :   $ **hexyl \--block-size=4kB \--length=20blocks**
```
sharkdp, line 73: Sorry. One more think. block**s** doesn't work, unfortunately. We should either make that work (I'd think it would be relatively easy) or change the man page here once more.

sharkdp (approved):
Thank you!

PR #154, opened 2022-04-22
Added `--plain` `--no-chars` and `--no-position` toggles

mkatychev:
New flags added:
```
OPTIONS:
    -p, --plain                 Display output with --no-chars, --no-position, and
                                --border=none.
    -C, --no-chars              Whether to display the character table on the right.
    -P, --no-position           Whether to display the position indicator on the left.
```

Motivation was mainly for diffing two separate binary files, borders could be removed but char table and position still created a lot of noise.

Output example:
image


src/lib.rs
```diff
@@ -179,16 +185,20 @@ impl<'a, Writer: Write> Printer<'a, Writer> {
...
-            byte_char_table: (0u8..=u8::max_value())
-                .map(|i| {
-                    let byte_char = format!("{}", Byte(i).as_char());
-                    if show_color {
-                        Byte(i).color().paint(byte_char).to_string()
-                    } else {
-                        byte_char
-                    }
+            byte_char_table: show_char_table
+                .then(|| {
```
mkatychev, line 189: This is failing `cargo clippy` on rust 1.46 (this was set on 2021-11-10, over 6 months ago), if the project is bound to that version I'm happy to whip up an alternative. The run-time cost of this allocation is pretty minimal regardless.

sharkdp: I'm fine with bumping the Rust version, if that provides any benefits here.

sharkdp:
Before I proceed to review this: how does this integrate with what is being proposed in #149?

mkatychev:
> Before I proceed to review this: how does this integrate with what is being proposed in #149?

This PR does not toggle separators or modify existing border behaviour, it allows the toggling of the character table (on the right) and the hex position (on the right).

* This PR allows one to disable the 1st or 3rd column
*  `--plain` disables both and applies the existing `--border=none` flag as a convenience flag

src/bin/hexyl.rs
```diff
@@ -104,24 +104,46 @@ fn run() -> Result<(), AnyhowError> {
...
                 .help(
                     "When to use colors. The auto-mode only displays colors if the output \
                      goes to an interactive terminal",
                 ),
         )
+        .arg(
+            Arg::with_name("plain")
+                .short("p")
+                .long("plain")
+                .help("Display output with --no-chars, --no-position, and --border=none."),
```
sharkdp, line 118: Should it be mentioned that colors are also disabled?

src/bin/hexyl.rs
```diff
@@ -104,24 +104,46 @@ fn run() -> Result<(), AnyhowError> {
...
+        )
         .arg(
             Arg::with_name("border")
                 .long("border")
                 .takes_value(true)
                 .value_name("STYLE")
                 .possible_values(&["unicode", "ascii", "none"])
+                .default_value_if("plain", None, "none")
```
sharkdp, line 126: Nice, didn't know about `default_value_if`.

src/bin/hexyl.rs
```diff
@@ -104,24 +104,46 @@ fn run() -> Result<(), AnyhowError> {
...
                 .default_value("unicode")
                 .help(
                     "Whether to draw a border with Unicode characters, ASCII characters, \
                     or none at all",
                 ),
         )
+        .arg(
+            Arg::with_name("no_chars")
+                .short("C")
+                .long("no-chars")
```
sharkdp, line 136: Is there any precedent for this option name in other hex viewers? Or for that terminology ("characters"/"chars")? If not, what would be alternatives?

I usually prefer the long options the be self-descriptive (and longer, if necessary). Something like `--no-character-panel`. Especially if we have shell completions and additional short options.

src/bin/hexyl.rs
```diff
@@ -104,24 +104,46 @@ fn run() -> Result<(), AnyhowError> {
...
+                .conflicts_with("plain")
```
sharkdp, line 137: Does this have to be conflicting with `--plain`? It's not really necessary, right?

src/bin/hexyl.rs
```diff
@@ -104,24 +104,46 @@ fn run() -> Result<(), AnyhowError> {
...
+                .help("Whether to display the character table on the right."),
+        )
+        .arg(
+            Arg::with_name("no_position")
```
sharkdp, line 141: Same questions here: do we have a similar flag in other hex viewers? If so, how is it called? If there is nothing similar, maybe `--no-file-position`?

src/bin/hexyl.rs
```diff
@@ -104,24 +104,46 @@ fn run() -> Result<(), AnyhowError> {
...
+                .short("P")
+                .long("no-position")
+                .conflicts_with("plain")
```
sharkdp, line 144: see above

src/bin/hexyl.rs
```diff
@@ -237,6 +259,13 @@ fn run() -> Result<(), AnyhowError> {
 
     let squeeze = !matches.is_present("nosqueezing");
 
+    // -nc true && false
+    // -plain true && false
```
sharkdp, line 263: What is this?

sharkdp (changes requested):
Nice! Thank you. Mainly have some questions regrading the command-line interface (option names)

src/bin/hexyl.rs
mkatychev: I largely referred to variable naming withing the code itself, in hindsight that might not be the best source.
I like `--no-character-panel` though I would propose that the suffix table be used: `--no-character-table` mainly because  `_table` is already used in the codebase:
https://github.com/sharkdp/hexyl/blob/a23645ddfdc1d26a64f2d6d6894ac577058df134/src/lib.rs#L151-L152

src/bin/hexyl.rs
mkatychev: I suppose not, though it would be good to avoid using them together since that's redundant.

src/bin/hexyl.rs
mkatychev: I'd suggest to avoid using the term file esp. when pipe operations like the one below are somewhat common (in my workflow) to use:
```
$ echo abcdef1234567890 | hexyl
│00000000│ 61 62 63 64 65 66 31 32 ┊ 33 34 35 36 37 38 39 30 │abcdef12┊34567890│
│00000010│ 0a                      ┊                         │_       ┊        │
```

`hexdump` seems to use a [formatter](https://stackoverflow.com/questions/2614764/how-to-create-a-hex-dump-of-file-containing-only-the-hex-characters-without-spac) over any toggling, which is fine but much more work:

```
$ echo 'abcdef1234567890' | hexdump -ve '1/1 "%.2x"'
616263646566313233343536373839300a
```

perhaps `--no-position-table` or `--no-byte-hex-table`?

src/bin/hexyl.rs
mkatychev: oversight

src/bin/hexyl.rs
sharkdp: True, it's redundant. But imagine someone typically wants to use `hexyl` without the char. table or the position and sets an alias:
```bash
alias hexyl="hexyl --no-file-position"
```

Then they would still be able to use `hexyl --plain` if they want to.

src/bin/hexyl.rs
sharkdp: Let's go with `--no-position` for now.

src/bin/hexyl.rs
sharkdp: Hm. I don't like the "table" suffix. Neither in the code, nor in the CLI. If at all, it's a single column of the table. I suggest we either use:
```
--no-position
--no-characters
```
or
```
--no-position-panel
--no-character-panel
```

I'm fine with both alternatives.

sharkdp (approved):
Thank you for the updates!

PR #161, opened 2022-06-03
Adds the --color-mode flag.

### LINKED ISSUE #156: 256-color still in use
#38 still left some use of 256-colors for gray; when using tmux on the Linux kernel console, this color gets quantitized to black, so addresses are unreadable

https://github.com/sharkdp/hexyl/blob/master/src/lib.rs#L15-L16=

remexre:
Closes #156.

This also adds an environment variable, `HEXYL_COLOR_MODE`, so that the color mode can be set by the end user when hexyl is used in scripts. (If you'd rather there be a config file or something instead, I'd be open to doing that instead.)


src/lib.rs
```diff
@@ -5,19 +5,25 @@ pub use input::*;
...
 
 const BUFFER_SIZE: usize = 256;
 
-const COLOR_NULL: Color = Fixed(242); // grey
-const COLOR_OFFSET: Color = Fixed(242); // grey
-const COLOR_ASCII_PRINTABLE: Color = Color::Cyan;
-const COLOR_ASCII_WHITESPACE: Color = Color::Green;
-const COLOR_ASCII_OTHER: Color = Color::Purple;
-const COLOR_NONASCII: Color = Color::Yellow;
+lazy_static::lazy_static! {
```
sharkdp, line 15: There is a more modern alternative to `lazy_static` called [`once_cell`](https://docs.rs/once_cell/latest/once_cell/). Maybe we could use it here? (See also: https://docs.rs/once_cell/latest/once_cell/#faq)

src/lib.rs
```diff
@@ -146,6 +158,7 @@ pub struct Printer<'a, Writer: Write> {
     buffer_line: Vec<u8>,
     writer: &'a mut Writer,
     show_color: bool,
+    use_8_bit_color: bool,
```
sharkdp, line 161: I would prefer if we use an enum instead of a bool. Maybe
```suggestion
enum ColorMode {
  Mode16,
  Mode8Bit,
}
```

Maybe it also makes sense to combine it with the `show_color` bool?

src/bin/hexyl.rs
```diff
@@ -109,6 +109,19 @@ fn run() -> Result<(), AnyhowError> {
...
                 ),
         )
+        .arg(
+            Arg::new("color-mode")
+                .long("color-mode")
+                .takes_value(true)
+                .value_name("MODE")
+                .possible_values(&["auto", "16", "8-bit"])
+                .default_value("auto")
+                .env("HEXYL_COLOR_MODE")
```
sharkdp, line 119: nice :+1:

src/bin/hexyl.rs
```diff
@@ -109,6 +109,19 @@ fn run() -> Result<(), AnyhowError> {
...
+                .help(
+                    "The set of colors to use. The auto-mode currently is equivalent to 8-bit, \
+                     but this may change in the future",
```
sharkdp, line 122: Should we document the `HEXYL_COLOR_MODE` variable or is that done by clap?

Maybe add a short explanation what "16" and "8-bit" means?

src/lib.rs
remexre: Done in 61aea39165a7a798f2ab6650f66cfdbf2887c514

src/lib.rs
```diff
@@ -5,19 +5,25 @@ pub use input::*;
...
+static STYLE_NULL_16: Lazy<Style> = Lazy::new(|| Color::Black.bold()); // grey
+static STYLE_OFFSET_16: Lazy<Style> = Lazy::new(|| Color::Black.bold()); // grey
+
+static STYLE_NULL_8BIT: Lazy<Style> = Lazy::new(|| Fixed(242).normal()); // grey
+static STYLE_OFFSET_8BIT: Lazy<Style> = Lazy::new(|| Fixed(242).normal()); // grey
+
+static STYLE_ASCII_PRINTABLE: Lazy<Style> = Lazy::new(|| Color::Cyan.normal());
+static STYLE_ASCII_WHITESPACE: Lazy<Style> = Lazy::new(|| Color::Green.normal());
+static STYLE_ASCII_OTHER: Lazy<Style> = Lazy::new(|| Color::Purple.normal());
+static STYLE_NONASCII: Lazy<Style> = Lazy::new(|| Color::Yellow.normal());
```
sharkdp, line 26: At some point we should probably refactor this into a single data structure behind one `Lazy<…>` construct.

Does this change (`const` => `Lazy`) introduce any performance penalty?

src/lib.rs
```diff
@@ -145,7 +163,7 @@ pub struct Printer<'a, Writer: Write> {
     /// The buffered line built with each byte, ready to print to writer.
     buffer_line: Vec<u8>,
     writer: &'a mut Writer,
-    show_color: bool,
+    color_mode: Option<ColorMode>,
```
sharkdp, line 166: Not necessary to change this now, but I was wondering if the code would improve here and there if we were to introduce `ColorMode::NoColor` instead of wrapping `ColorMode` inside an `Option<…>`.

sharkdp (approved):
Code changes look great. I added a couple of optional comments. And one regarding performance, which I would like to clarify before we merge this.

Issue #66, opened 2019-07-09
Beat xxd


mike239x:
I did a bit of benchmarking and I can't help but notice that `xxd` is faster than `hexyl`.
On my machine on a file of about 700M:
```
$ time xxd myfile > /dev/null

real	0m43.245s
user	0m42.950s
sys	0m0.272s

$ time hexyl --color=never --no-squeezing --border=none myfile > /dev/null

real	1m10.967s
user	1m1.371s
sys	0m9.592s
```

It would be nice to beat `xxd` in speed... I got no idea how to do it though.

sharkdp:
Thank you for the feedback.

I agree, it would be nice. But not more. I don't really see a problem with the current speed, as I don't think that performance (at the current level) is critical for a hexdump tool. `hexyl` processes around 10 MiB of data per second. It outputs text to the screen at a speed that is much faster than terminal emulators can handle (in terminator, `hexyl` is a factor of 5 slower when I write to the TTY).

In which real-world use case would we really need it to be faster?

mike239x:
I tried to find a "real-world use case" but failed. I would say it is an ideological thing... Something in lines "software shouldn't get slower with time, but faster".

I'll take a look into the source code in my free time, maybe (though unlikely) I'll find the way to improve it :)

sharkdp:
> "software shouldn't get slower with time, but faster"

I would agree. But `hexyl` is about adding additional functionality (the colorized output). It's not trying to be a 1:1 replacement for `xxd`.

remexre:
Real-world use case -- I've got a pretty big file that's mostly zeroes, with a k or so of nonzero data. Reading from /tmp, `hexyl` takes 55.791s, `hexdump -C` takes 1.091s, `xxd >/dev/null` takes 40.528s.

sharkdp:
@remexre Thank you.

If someone wants to work on this, here is a reproducible benchmark:
```bash
#!/bin/bash

dd if=/dev/zero    bs=10M count=1 >  data
dd if=/dev/urandom bs=1k  count=1 >> data

hyperfine --warmup 3 \
    'hexyl data' \
    'hexyl --no-squeezing data' \
    'hexdump -C data' \
    'hexdump -C --no-squeezing data' \
    'xxd data' \
    --export-markdown results.md
```

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
[...]

PR #164, opened 2022-07-12
Adds --columns for Print using custom width #13

### LINKED ISSUE #13: Print using custom width
This is excellent, and the default width of 16 is great! It'd be really nice if we could print according to custom alignments, though -- this is a fairly standard feature in most hex editors, and I can't imagine it being egregiously complex to add this.

Another couple of ideas:
* Getting the max width based on a *provided* terminal width would be nice -- this would let me simply plug in a width and let `hexyl` calculate the space overhead of borders for me.
* An auto-width flag (which is like the above for the current terminal width) to be used would also be handy -- it would have some overlap with the above, but there could be reasons for setting

Prototypical suggestion for options:

```
# Ideas for a fixed column width:
-c --columns        Sets the number of hex data columns to be displayed.
                    Cannot be used with other width-setting options.
-w --width

# Ideas for terminal auto-width:
-a --auto-width     Sets the number of hex data columns to be adjusted according
                    to the detected terminal width. Conceptually, this could be an
                    alias of `-t $terminal_width`.
                    Cannot be used with other width-setting options.

# Ideas for fixed terminal width:
-t --terminal-width Sets the number of terminal columns to be displayed.
                    Since the terminal width may not be an evenly divisible by the 
                    width per hex data column, this will use the greatest number of 
                    hex data columns that can fit in the requested width but still
                    leave some space to the right.
                    Cannot be used with other width-setting options.
```

I'm more than happy to push on this, if effort is a concern. :)

sharifhsn:
This pull request fixes #13 partially by adding the `--columns` option to manually set the number of columns. The default is 2 as with previous behavior.

Some notes:
- the short option is `-w` not `-c` as that option is already reserved for `--bytes`
- it passes `cargo test`
- I could have added the option to automatically set columns to terminal width, but that would require adding another dependency

### DIFF
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -4,7 +4,7 @@ extern crate clap;
 use std::convert::TryFrom;
 use std::fs::File;
 use std::io::{self, prelude::*, SeekFrom};
-use std::num::{NonZeroI64, NonZeroU8};
+use std::num::{NonZeroI64, NonZeroU16, NonZeroU8};
 
 use clap::{crate_name, crate_version, AppSettings, Arg, ColorChoice, Command};
 
@@ -15,6 +15,8 @@ use anyhow::{anyhow, Context, Error as AnyhowError};
 use const_format::formatcp;
 
 use thiserror::Error as ThisError;
+
+use terminal_size::terminal_size;
 
 use hexyl::{BorderStyle, Input, Printer};
 
@@ -160,6 +162,31 @@ fn run() -> Result<(), AnyhowError> {
                     "Sets the number of hex data columns to be displayed. \
                     Cannot be used with other width-setting options.",
                 ),
+        )
+        .arg(
+            Arg::new("terminal_width")
+                .short('t')
+                .long("terminal-width")
+                .takes_value(true)
+                .value_name("N")
+                .conflicts_with("columns")
+                .help(
+                    "Sets the number of terminal columns to be displayed.\nSince the terminal \
+                    width may not be an evenly divisible by the width per hex data column, this \
+                    will use the greatest number of hex data columns that can fit in the requested \
+                    width but still leave some space to the right.\nCannot be used with other \
+                    width-setting options.",
+                ),
+        )
+        .arg(
+            Arg::new("auto_width")
+                .short('a')
+                .long("auto-width")
+                .conflicts_with_all(&["columns", "terminal_width"])
+                .help(
+                    "Sets the number of hex data columns to be adjusted according to the \
+                    detected terminal width.\nCannot be used with other width-setting options.",
+                ),
         );
 
     let matches = command.get_matches();
@@ -278,7 +305,7 @@ fn run() -> Result<(), AnyhowError> {
         .transpose()?
         .unwrap_or(0);
 
-    let columns = matches
+    let columns = if let Some(columns) = matches
         .value_of("columns")
         .map(|s| {
             s.parse::<NonZeroU8>().map(u8::from).context(anyhow!(
@@ -287,7 +314,40 @@ fn run() -> Result<(), AnyhowError> {
             ))
         })
         .transpose()?
-        .unwrap_or(2);
+    {
+        columns
+    } else {
+        let terminal_width = if let Some(terminal_width) = matches
+            .value_of("terminal_width")
+            .map(|s| {
+                s.parse::<NonZeroU16>().map(u16::from).context(anyhow!(
+                    "failed to parse `--columns` arg {:?} as unsigned nonzero integer",
+                    s
+                ))
+            })
+            .transpose()?
+        {
+            Some(terminal_width)
+        } else if matches.is_present("auto_width") {
+            Some(terminal_size().expect("not a tty").0 .0)
+        } else {
+            None
+        };
+
+        if let Some(terminal_width) = terminal_width {
+            let offset = if show_position_panel { 10 } else { 1 };
+            let col_width = if show_char_panel { 35 } else { 26 };
+            if (terminal_width - offset) / col_width < 1 {
+                1
+            } else {
+                ((terminal_width - offset) / col_width)
+                    .try_into()
+                    .expect("there is a maximum of 255 columns")
+            }
+        } else {
+            2
+        }
+    };
 
     let stdout = io::stdout();
     let mut stdout_lock = stdout.lock();
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -160,7 +160,8 @@ fn run() -> Result<(), AnyhowError> {
                 .value_name("N")
                 .help(
                     "Sets the number of hex data columns to be displayed. \
-                    Cannot be used with other width-setting options.",
+                    `--columns=auto` will display the maximum number of hex data columns \
+                    based on the current terminal width",
                 ),
         )
         .arg(
@@ -176,16 +177,6 @@ fn run() -> Result<(), AnyhowError> {
                     will use the greatest number of hex data columns that can fit in the requested \
                     width but still leave some space to the right.\nCannot be used with other \
                     width-setting options.",
-                ),
-        )
-        .arg(
-            Arg::new("auto_width")
-                .short('a')
-                .long("auto-width")
-                .conflicts_with_all(&["columns", "terminal_width"])
-                .help(
-                    "Sets the number of hex data columns to be adjusted according to the \
-                    detected terminal width.\nCannot be used with other width-setting options.",
                 ),
         );
 
@@ -305,7 +296,19 @@ fn run() -> Result<(), AnyhowError> {
         .transpose()?
         .unwrap_or(0);
 
-    let columns = if let Some(columns) = matches
+    let max_columns_fn = |terminal_width: u16| {
+        let offset = if show_position_panel { 10 } else { 1 };
+        let col_width = if show_char_panel { 35 } else { 26 };
+        if (terminal_width - offset) / col_width < 1 {
+            1
+        } else {
+            (terminal_width - offset) / col_width
+        }
+    };
+
+    let columns = if matches.value_of("columns") == Some("auto") {
+        max_columns_fn(terminal_size().ok_or_else(|| anyhow!("not a TTY"))?.0 .0)
+    } else if let Some(columns) = matches
         .value_of("columns")
         .map(|s| {
             s.parse::<NonZeroU16>().map(u16::from).context(anyhow!(
@@ -316,8 +319,7 @@ fn run() -> Result<(), AnyhowError> {
         .transpose()?
     {
         columns
-    } else {
-        let terminal_width = if let Some(terminal_width) = matches
+    } else if let Some(terminal_width) = matches
             .value_of("terminal_width")
             .map(|s| {
                 s.parse::<NonZeroU16>().map(u16::from).context(anyhow!(
@@ -327,24 +329,9 @@ fn run() -> Result<(), AnyhowError> {
             })
             .transpose()?
         {
-            Some(terminal_width)
-        } else if matches.is_present("auto_width") {
-            Some(terminal_size().ok_or_else(|| anyhow!("not a TTY"))?.0 .0)
-        } else {
-            None
-        };
-
-        if let Some(terminal_width) = terminal_width {
-            let offset = if show_position_panel { 10 } else { 1 };
-            let col_width = if show_char_panel { 35 } else { 26 };
-            if (terminal_width - offset) / col_width < 1 {
-                1
-            } else {
-                (terminal_width - offset) / col_width
-            }
+        max_columns_fn(terminal_width)
         } else {
             2
-        }
     };
 
     let stdout = io::stdout();
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -154,7 +154,7 @@ fn run() -> Result<(), AnyhowError> {
         )
         .arg(
             Arg::new("panels")
-                .short('p')
+                .short('w')
                 .long("panels")
                 .takes_value(true)
                 .value_name("N")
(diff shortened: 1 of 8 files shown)


sharifhsn:
I have a pull request ready for terminal width, but it requires adding the `terminal_width` dependency. I'll wait to submit it until this request is approved.

src/bin/hexyl.rs
```diff
@@ -149,6 +149,17 @@ fn run() -> Result<(), AnyhowError> {
...
+        )
+        .arg(
+            Arg::new("columns")
+                .short('w')
+                .long("columns")
+                .takes_value(true)
+                .value_name("N")
+                .help(
+                    "Sets the number of hex data columns to be displayed. \
+                    Cannot be used with other width-setting options.",
```
sharkdp, line 161: Should this be omitted for now?

sharifhsn: I already have the pull request for terminal width ready, I'll commit it if there's no other problems. I just wanted to make sure that the major architectural changes were correct.

src/squeezer.rs
```diff
@@ -111,13 +110,14 @@ impl Squeezer {
 mod tests {
     use super::*;
 
-    const LSIZE_USIZE: usize = LSIZE as usize;
+    const LSIZE: u64 = 16;
+    const LSIZE_USIZE: usize = 16;
```
sharkdp, line 114: Maybe
```suggestion
    const LSIZE: u64 = 16;
    const LSIZE_USIZE: usize = LSIZE as usize;
```

sharkdp: Or just use `LSIZE` and use `LSIZE as usize` in a few places below?

sharkdp (commented):
Thank you very much. Works great!

Does this have any impact on performance? (see e.g. https://github.com/sharkdp/hexyl/issues/66#issuecomment-511125330 on how to perform `hexyl` benchmarks using `hyperfine`)

sharifhsn:
> Thank you very much. Works great!
> 
> Does this have any impact on performance? (see e.g. [#66 (comment)](https://github.com/sharkdp/hexyl/issues/66#issuecomment-511125330) on how to perform `hexyl` benchmarks using `hyperfine`)

No, I'll do that now. I couldn't find a benchmark harness before, thank you for pointing me to it.

sharkdp:
We should probably add a script with a few useful benchmarks, true. For this PR, I would suggest to use something like:

```bash
#!/bin/bash

dd if=/dev/urandom bs=1M count=1 of=/tmp/data

hyperfine \
    --warmup 3 \
    --shell=none \
    --output=pipe \
    --export-markdown=result-pipe.md \
    'hexyl /tmp/data'

hyperfine \
    --warmup 3 \
    --shell=none \
    --output=inherit \
    --export-markdown=result-inherit.md \
    'hexyl /tmp/data'

rm /tmp/data
```

Here, we use `hyperfine`s new `--output` option to specify where to redirect the output of the benchmarked program. Using "pipe" is like doing `hexyl … | cat > /dev/null` (which is much different from doing `hexyl … > /dev/null` because `hexyl` doesn't directly write to `/dev/null` and kernel optimizations don't apply, i.e. `write` calls actually need to take place). And using "inherit" is like calling the program with the TTY attached (this will output a lot to your screen).

sharkdp:
You can also create two binaries, `hexyl-master` and `hexyl-feature`, and then run something like:
```
hyperfine \
    --warmup 3 \
    --shell=none \
    --output=pipe \
    --parameter-list version master,feature
    --export-markdown=result-pipe.md \
    './hexyl-{version} /tmp/data'
```
to see the direct comparison.

sharifhsn:
|:---|---:|---:|---:|---:|
| `hexyl data` | 421.3 ± 14.3 | 404.5 | 444.0 | 1.24 ± 0.11 |
| `hexyl --no-squeezing data` | 563.5 ± 14.7 | 542.3 | 583.4 | 1.65 ± 0.15 |
| `target/release/hexyl data` | 433.0 ± 16.0 | 409.1 | 462.5 | 1.27 ± 0.12 |
| `target/release/hexyl --no-squeezing data` | 526.2 ± 18.1 | 506.6 | 559.6 | 1.55 ± 0.14 |
| `target/release/hexyl --columns 4 data` | 367.4 ± 12.7 | 347.8 | 382.6 | 1.08 ± 0.10 |
| `target/release/hexyl --columns 4 --no-squeezing data` | 407.6 ± 14.2 | 375.5 | 426.5 | 1.20 ± 0.11 |
| `target/release/hexyl --columns 8 data` | 340.5 ± 28.6 | 306.6 | 399.2 | 1.00 |
| `target/release/hexyl --columns 8 --no-squeezing data` | 351.9 ± 15.3 | 318.9 | 366.2 | 1.03 ± 0.10 |

These are my results using 1 MB of `/dev/urandom`. I used my system-installed binary of `hexyl` to compare since I'm pretty sure there were no feature changes since the last release.

It seems as though for the default width of two columns the new version is slightly slower, but within noise for the benchmark. With more columns it's much faster, but that of course makes sense since it's writing fewer lines.

sharifhsn:
One more thing; I've implemented `columns` here as a `u8`, since I find it unlikely that anyone would ever use over 255 columns in the viewer. This also places a cap on terminal width at 8969. Is this desirable, or should I change the type to `u16`?

src/bin/hexyl.rs
sharkdp: Ok, in that case, yes. Please add it to this PR. Or leave the help text as-is, and we do a second PR with the auto-width changes.

sharifhsn:
I've added the terminal width setting, so this should completely close #13.

sharkdp:
> One more thing; I've implemented `columns` here as a `u8`, since I find it unlikely that anyone would ever use over 255 columns in the viewer. This also places a cap on terminal width at 8969. Is this desirable, or should I change the type to `u16`?

I wanted to remark this in my first review but didn't want to be overly annoying :smile:. In my personal experience: trying to use size-limited (unsigned) integers (u8, u16, i8, i16) in places which are *not* size- or performance critical is almost always a mistake. It's usually meant as a (premature) optimization. But the reality is that it (1) leads to more errors due to overflows (2) lead to more complicated code due to more explicit casts between 8-bit, 16-bit and 32-bit numbers and (3) don't even necessarily generate faster code (because if they are not closely packed, we can't really make use of them) or smaller data structs (due to alignment mismatches).

So personally, for a single parameter, I would never mind trying to go for a small integer type and simply go with a 32-bit integer, or even `isize`/`usize`. Often, it's even undesirable to use unsigned integers for parameters that cannot (reasonably) be negative, because again, it might lead to more complicated code, and it can lead to tricky overflow errors.

sharkdp:
Feel free to bump the min. supported Rust version if that causes problems.

sharkdp:
Concerning the newly added options:

Would it be possible to scratch `--auto-width` and use something like `--columns=max` or `--columns=auto` instead?

Also, I think `--terminal-width` should be usable with other "width-setting" options. I think of it purely as an overwrite for the auto-detected terminal width. What do you think?

> I've added the terminal width setting, so this should completely close #13.

Does it? I believe the issue also asks for custom widths that would not be multiples of 16. Like if you have some data structure with elements that are 14 byte long. Then you might want to specify `--width=14`.

sharifhsn:
> > One more thing; I've implemented `columns` here as a `u8`, since I find it unlikely that anyone would ever use over 255 columns in the viewer. This also places a cap on terminal width at 8969. Is this desirable, or should I change the type to `u16`?
> 
> I wanted to remark this in my first review but didn't want to be overly annoying smile. In my personal experience: trying to use size-limited (unsigned) integers (u8, u16, i8, i16) in places which are _not_ size- or performance critical is almost always a mistake. It's usually meant as a (premature) optimization. But the reality is that it (1) leads to more errors due to overflows (2) leads to more complicated code due to more explicit casts between 8-bit, 16-bit and 32-bit numbers and (3) doesn't even necessarily generate faster code (because if they are not closely packed, we can't really make use of them) or smaller data structs (due to alignment mismatches).
> 
> So personally, for a single parameter, I would never mind trying to go for a small integer type and simply go with a 32-bit integer, or even `isize`/`usize`. Often, it's even undesirable to use unsigned integers for parameters that cannot (reasonably) be negative, because again, it might lead to more complicated code, and it can lead to tricky overflow errors.

Thank you for saying this! I am quite new to Rust so I appreciate the advice. Making it a `u16` eliminates a cast (since `terminal_width` uses a `u16`) so I'll use that instead. Using a bigger value would be more of a hassle than a `u16`.

src/bin/hexyl.rs
```diff
@@ -267,6 +305,50 @@ fn run() -> Result<(), AnyhowError> {
...
+                s.parse::<NonZeroU16>().map(u16::from).context(anyhow!(
+                    "failed to parse `--columns` arg {:?} as unsigned nonzero integer",
+                    s
+                ))
+            })
+            .transpose()?
+        {
+            Some(terminal_width)
+        } else if matches.is_present("auto_width") {
+            Some(terminal_size().expect("not a tty").0 .0)
```
sharkdp, line 332: I think we should properly handle this case (not a tty), by which I mean: return a proper error here and show a nice error message for the user. This can easily be triggered if the output of hexyl is being redirected.

src/lib.rs
```diff
@@ -164,6 +166,7 @@ impl<'a, Writer: Write> Printer<'a, Writer> {
         show_position_panel: bool,
         border_style: BorderStyle,
         use_squeeze: bool,
+        columns: u8,
```
sharkdp, line 169: This is a breaking change for `hexyl`-as-a-library. This whole `new` method is really unfortunate. There are too many arguments even before this PR. We should really use a builder-pattern instead.

But that's not really the problem of your PR.

src/lib.rs
```diff
@@ -347,45 +351,57 @@ impl<'a, Writer: Write> Printer<'a, Writer> {
...
-                let _ = write!(
-                    &mut self.buffer_line,
-                    "{0:1$}{2}",
-                    "",
-                    3 * (16 - len),
-                    self.border_style.outer_sep()
-                );
+            if len < usize::from(8 * self.columns) {
+                write!(&mut self.buffer_line, "{0:1$}", "", 3 * (8 - len % 8))?;
+                // dbg!(usize::from(8 * self.columns) - (len + (8 - len % 8)));
```
sharkdp, line 392: remove?

sharifhsn: Ah, my sloppiness strikes again. Good eye.

sharifhsn:
> Concerning the newly added options:
> 
> Would it be possible to scratch `--auto-width` and use something like `--columns=max` or `--columns=auto` instead?

Of course.

> Also, I think `--terminal-width` should be usable with other "width-setting" options. I think of it purely as an overwrite for the auto-detected terminal width. What do you think?

I was only going off of the original issue's suggestions. The use case for `--terminal-width` is probably in a situation where the width is necessarily limited for some reason. In that case, I would pick the minimum between the two options i.e. prefer `--columns` unless the number of columns exceeds that which is dictated by `--terminal-width`.

> > I've added the terminal width setting, so this should completely close #13.
> 
> Does it? I believe the issue also asks for custom widths that would not be multiples of 16. Like if you have some data structure with elements that are 14 byte long. Then you might want to specify `--width=14`.

Ah, I wasn't aware that was in the scope of the original issue. That would probably require a more significant rewrite.

src/lib.rs
sharifhsn: I would be willing to work on implementing a builder pattern if you're committed to pushing a breaking change anyway.

sharkdp:
> Ah, I wasn't aware that was in the scope of the original issue. That would probably require a more significant rewrite.

That's okay. Let's focus on `--columns` for now.

sharkdp:
Another option would be to leave the option name as is, but change the meaning. And then for this PR, we only allow `--columns=8,16,24,…`. What do you think?

sharifhsn:
> Another option would be to leave the option name as is, but change the meaning. And then for this PR, we only allow `--columns=8,16,24,…`. What do you think?

I think it would be more ergonomic here to have the two options, columns-per-panel and number-of-panels.

In my mind, the idea behind variable number columns is to have the panel size itself adjust. With the other option, the panel size would always be 8 bytes, and a future ability to have e.g. `columns=14` would leave blank space at the end.

It would be better to set `columns=7` and `panels=2` which is more explicit in my opinion.

sharkdp:
You are right.

Let's change the command line option name then 👍🏼

src/bin/hexyl.rs
```diff
@@ -149,6 +151,33 @@ fn run() -> Result<(), AnyhowError> {
...
+                .value_name("N")
+                .help(
+                    "Sets the number of hex data panels to be displayed. \
+                    `--panels=auto` will display the maximum number of hex data panels \
+                    based on the current terminal width",
+                ),
+        )
+        .arg(
+            Arg::new("terminal_width")
+                .short('t')
```
sharkdp, line 169: I'm trying to be conservative with short options. There are only a few of them :-). I would suggest that we remove both short options for now (i.e.: same for `-w`).

sharkdp:
> I have a commit ready for a builder pattern. It uses a new PrinterBuilder struct, so it can technically be implemented in a non-breaking way, although that's probably undesirable. It's quite out of scope for this pull request, so I'll submit a new one once this one is approved.

That would be great! If we do that, I would make that a breaking change, i.e. remove the old `new` method.

> @sharkdp I think we can declare this PR complete? I don't see any other roadblocks.

Yes. Looks great. Two things: the inline comment regarding the short options. And it would be great if you could add a CHANGELOG entry in the "unreleased" section. The format for entries is:
```
- Description what has been changed, see #123 (@user)
```
where `#123` links to the bug ticket and/or the PR and `user` is your username.

sharkdp:
Thank you!

PR #168, opened 2022-09-17
Implement builder pattern

sharifhsn:
As discussed in my previous pull request, this pull request deprecates using the new method directly on the `Printer` struct, preferring the `PrinterBuilder` struct for the builder pattern. This is a breaking change for `hexyl` as a library.

### DIFF
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -1,6 +1,6 @@
 use std::io;
 
-use hexyl::{BorderStyle, Printer};
+use hexyl::{BorderStyle, PrinterBuilder};
 
 fn main() {
     let input = vec![
@@ -16,16 +16,15 @@ fn main() {
     let show_position_panel = true;
     let use_squeezing = false;
     let border_style = BorderStyle::Unicode;
-    let columns = 2;
+    let panels = 2;
 
-    let mut printer = Printer::new(
-        &mut handle,
-        show_color,
-        show_char_panel,
-        show_position_panel,
-        border_style,
-        use_squeezing,
-        columns,
-    );
+    let mut printer = PrinterBuilder::new(&mut handle)
+        .show_color(show_color)
+        .show_char_panel(show_char_panel)
+        .show_position_panel(show_position_panel)
+        .with_border_style(border_style)
+        .with_squeeze(use_squeezing)
+        .with_panels(panels)
+        .build();
     printer.print_all(&input[..]).unwrap();
 }
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -16,7 +16,7 @@ fn main() {
         .show_char_panel(true)
         .show_position_panel(true)
         .with_border_style(BorderStyle::Unicode)
-        .with_squeeze(false)
+        .enable_squeezing(false)
         .with_panels(2)
         .build();
     printer.print_all(&input[..]).unwrap();
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -11,20 +11,13 @@ fn main() {
     let stdout = io::stdout();
     let mut handle = stdout.lock();
 
-    let show_color = true;
-    let show_char_panel = true;
-    let show_position_panel = true;
-    let use_squeezing = false;
-    let border_style = BorderStyle::Unicode;
-    let panels = 2;
-
     let mut printer = PrinterBuilder::new(&mut handle)
-        .show_color(show_color)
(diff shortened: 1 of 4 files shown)


sharifhsn:
Let me know if there are any bikeshed quibbles with the naming of methods and I'll be happy to change them!

sharkdp:
Awesome, thank you very much!

Could you please add an entry to the "unreleased" section in `CHANGELOG.md` to briefly describe the breaking change? That would help me with creating the next release.

src/lib.rs
```diff
@@ -138,6 +139,77 @@ impl BorderStyle {
...
+        self.show_position_panel = show_position_panel;
+        self
+    }
+
+    pub fn with_border_style(mut self, border_style: BorderStyle) -> Self {
+        self.border_style = border_style;
+        self
+    }
+
+    pub fn with_squeeze(mut self, use_squeeze: bool) -> Self {
```
sharkdp, line 190: Maybe "enable squeezing"?
```suggestion
    pub fn enable_squeezing(mut self, enable: bool) -> Self {
```

src/lib.rs
```diff
@@ -138,6 +139,77 @@ impl BorderStyle {
...
+            border_style: BorderStyle::Unicode,
+            use_squeeze: true,
+            panels: 2,
+        }
+    }
+
+    pub fn with_writer(mut self, writer: &'a mut Writer) -> PrinterBuilder<'a, Writer> {
+        self.writer = writer;
+        self
+    }
```
sharkdp, line 168: Why is this needed?

src/lib.rs
```diff
@@ -138,6 +139,77 @@ impl BorderStyle {
...
+        self.border_style = border_style;
+        self
+    }
+
+    pub fn with_squeeze(mut self, use_squeeze: bool) -> Self {
+        self.use_squeeze = use_squeeze;
+        self
+    }
+
+    pub fn with_panels(mut self, panels: u16) -> Self {
```
sharkdp, line 195: Maybe
```suggestion
    pub fn with_num_panels(mut self, panels: u16) -> Self {
```
? Not sure..

examples/simple.rs
```diff
@@ -16,16 +16,15 @@ fn main() {
...
-        columns,
-    );
+    let mut printer = PrinterBuilder::new(&mut handle)
+        .show_color(show_color)
+        .show_char_panel(show_char_panel)
+        .show_position_panel(show_position_panel)
+        .with_border_style(border_style)
+        .with_squeeze(use_squeezing)
+        .with_panels(panels)
+        .build();
```
sharkdp, line 28: Nice! I think we should take advantage of the self-documenting property of the new builder API and change this example... directly passing those values above into the methods, without using intermediate variables.

sharifhsn: Great idea!

src/lib.rs
sharifhsn: Leftover from a previous pattern where the writer was optional. Thanks for noticing it!

src/lib.rs
sharifhsn: 👍

src/lib.rs
sharifhsn: I think changing the argument of `panels` to `num` would have a similar effect. Better still, rename the function to simply `num_panels`. But I don't feel strongly about this, so if you feel that your suggestion is better I'll implement it.

sharifhsn:
One last thing: do you feel it's necessary to add documentation to the new `PrinterBuilder` struct? I feel like the builder functions are self-documenting, but perhaps the `new` function and `build` function should have examples for those unfamiliar with the builder pattern.

sharkdp:
> One last thing: do you feel it's necessary to add documentation to the new PrinterBuilder struct? I feel like the builder functions are self-documenting, but perhaps the new function and build function should have examples for those unfamiliar with the builder pattern.

Thank you for your concern. Adding documentation is always great, but I wouldn't see this as a requirement for this PR. This changeset already improves hexyl as a library significantly.

There was no documentation before because - to be honest - I only support hexyl-the-public-library in a "best effort" manner.

So I suggest we merge this as is, but new PRs are always welcome of course!

src/lib.rs
sharkdp: I like the `num_panels` idea!

sharkdp (approved):
Thank you!

PR #169, opened 2022-10-01
Changed null char to `•`

### LINKED ISSUE #166: Add option to disable printing 0 bytes as ascii 0.
Consider the following output

Capture

The right column present ASCII values of the hex column.

In the case of null bytes (value 0), they are printed as ascii 0 in the right column.

This is confusing to say the least. Consider for example the case of no color (it text after all. User will copy & paste it elsewhere):

Capture

As a comparison, here is xxd output, where they denote non-printable characters with a period instead:

Capture

I did not find any command line arguments to disable this.

I would like to request a mode of hexyl to print non-printable characters in a similar manner as xxd does it. Specifically I would like it to not insert printable ascii that does not belong in the ascii printout:

* 0
* _
* •
And possibly more.

LemonjamesD:
Kinda Resolves #166 
This PR will changed the null char to `•` but it will not add the option to make it that or change it back to `0`.

(For Hacktoberfest)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -61,7 +61,7 @@ impl Byte {
         use crate::ByteCategory::*;
 
         match self.category() {
-            Null => '0',
+            Null => '•',
             AsciiPrintable => self.0 as char,
             AsciiWhitespace if self.0 == 0x20 => ' ',
             AsciiWhitespace => '_',


sharkdp:
We already use a similar (the same?) Unicode code point for `AsciiOther` (see a few lines below your change), so that's not the best solution here, IMO.

sharkdp:
If you look a few lines below your change, you could see that we already use that sign for `AsciiOther`. I looked around a bit and tried a few Unicode characters. So far, I like `⋄` best. It can be recognized as a `0` and it's not an ASCII character.

LemonjamesD:
That would work, I'll see about changing it tonight if you're not gonna do it today

sharkdp:
Solved in 7d38f70

Issue #104, opened 2020-10-23
Other sizes of data (group size and Endianness)


ACleverDisguise:
I frequently have to dump data files (ADC output, for example) that don't just have byte-oriented data.  It would be nice to be able to specify data width in the dump so I get the hex data grouped in the natural data size instead of having to do the little-endian two-step and mentally group indistinguishable bytes by 2 or 4 or whatever.  Something like:

--word-size=1 _(uint8_t, default)_
--word-size=2 _(uint16_t)_
--word-size=4 _(uint32_t)_
--word-size=8 _(uint64_t)_
--word-size=16 _(uint128_t)_

That covers the common-ish types.  If you want to be really brave you could do weird crap like 3-byte or 17 byte, but that is likely low return on investment.

Not all such data is little-endian, so an extra flag for those cases where word-size > 1 would be:

--little-endian _(default)_
--big-endian

Also, interpretation could be signed or unsigned

--signed
--unsigned _(default)_

Of course with this you'd drop the byte-oriented colouration (but maybe with --signed you'd highlight negative numbers in red or something).

sharkdp:
Thank you for the feedback.

It's not entirely clear to me what the output would look like.

Say I choose `--word-size=2` (uint16_t) and the input contains `0xAB 0xCD 0x12 0x34`. Would you like to see
```
CDAB 3412
```
for `--little-endian`?

ACleverDisguise:
That's pretty much exactly what I was picturing, yes.

sharkdp:
This looks similar to `xxd`s `-groupsize` option if I am not mistaking:
```
       -g bytes | -groupsize bytes
              Separate the output of every <bytes> bytes (two hex characters or  eight
              bit-digits  each)  by  a whitespace.  Specify -g 0 to suppress grouping.
              <Bytes> defaults to 2 in normal mode, 4 in little-endian mode and  1  in
              bits mode.  Grouping does not apply to postscript or include style.
```

I recently came across this when reading [this blog post](https://fasterthanli.me/series/making-our-own-executable-packer/part-1) which makes use of `-g` to inspect ELF64 executables.

ACleverDisguise:
[...]

PR #170, opened 2022-10-04
Support byte grouping

RinHizakura:
This PR is going to support different sizes of data which can almost fit the requirement of #104, without the endian option. Since there's some "magic" number (e.g. the number "25") related to this originally, I also try my best to replace those with a specific abstraction so we can simpler add more features accordingly.  To be simple, also note that I make the panel size change dynamically according to the bytes in the group, so the string `No content to print` has to be modified to `No content` to fit the panel when `-g 16`. It's fine if you have better ideas about this change. 

If you think this PR makes sense, I will also add more test cases to robust this feature, thanks!


sharkdp:
Thank you for your PR. We recently merged #173 which introduced some large scale changes to the code, so there are a lot of conflicts now. Sorry for that.

RinHizakura:
I just fix the conflicts. Please take a look again!

sharkdp:
@mkatychev Maybe we should merge this first, what do you think?

sharifhsn:
Yes, that's a good idea. I can modify my `clap4` PR so that it updates this as well. Alternatively, `clap` could be updated and this PR would change to use it.

src/bin/hexyl.rs
```diff
@@ -163,6 +163,17 @@ fn run() -> Result<(), AnyhowError> {
                     based on the current terminal width",
                 ),
         )
+        .arg(
+            Arg::new("group_bytes")
+                .short('g')
+                .takes_value(true)
```
sharifhsn, line 169: Replace this option with `.num_args(1)`

src/bin/hexyl.rs
```diff
@@ -332,6 +343,25 @@ fn run() -> Result<(), AnyhowError> {
         2
     };
 
+    let group_bytes = if let Some(group_bytes) = matches
+        .value_of("group_bytes")
```
sharifhsn, line 347: Replace with `.get_one::<String>("group_bytes")`

sharifhsn (changes requested):
These are the only changes that are needed to update for `clap` 4.

RinHizakura:
Hi, some of the integration tests are provided now. Please check whether we'll also need more special patterns.

src/bin/hexyl.rs
```diff
@@ -166,6 +166,17 @@ fn run() -> Result<(), AnyhowError> {
...
         )
+        .arg(
+            Arg::new("group_bytes")
+                .short('g')
+                .num_args(1)
+                .value_name("N")
+                .help(
+                    "Sets the number of octets per group to be displayed. \
+                    The possible option will be 1, 2, 4, 8, otherwise it will be set \
+                    to 1 by default.",
```
sharkdp, line 177:
```suggestion
                    The possible options are 1, 2, 4 and 8. The default is 1.
```

src/bin/hexyl.rs
```diff
@@ -337,6 +348,25 @@ fn run() -> Result<(), AnyhowError> {
...
+                s
+            ))
+        })
+        .transpose()?
+    {
+        if (group_bytes <= 8) && ((group_bytes & (group_bytes - 1)) == 0) {
+            group_bytes
+        } else {
+            1
+        }
```
sharkdp, line 365: I think I would prefer a hard error (in case someone uses an illegal value for `group_bytes`), rather than a silent fallback to 1. Or why did you decide to implement it this waay?

tests/integration_tests.rs
```diff
@@ -271,3 +271,79 @@ mod display_settings {
...
+        hexyl()
+            .arg("ascii")
+            .arg("--color=never")
+            .arg("-g=2")
+            .assert()
+            .success()
+            .stdout(
+                "┌────────┬─────────────────────┬─────────────────────┬────────┬────────┐\n\
+             │00000000│ 3031 3233 3435 3637 ┊ 3839 6162 6364 650a │01234567┊89abcde_│\n\
+             └────────┴─────────────────────┴─────────────────────┴────────┴────────┘\n",
```
sharkdp, line 289: is it possible to format this in the "right" way? or does `cargo fmt` destroy it?

src/bin/hexyl.rs
```diff
@@ -166,6 +166,17 @@ fn run() -> Result<(), AnyhowError> {
                     based on the current terminal width",
                 ),
         )
+        .arg(
+            Arg::new("group_bytes")
+                .short('g')
```
sharkdp, line 171: Can we please add a long option for this (`--group-bytes`) and use it in the unit tests?

sharkdp (changes requested):
Thank you very much. This looks great, except for a few minor comments.

sharkdp (approved):
Thank you

PR #175, opened 2022-11-16
Update to `clap` 4

sharifhsn:
The `clap` dependency is updated to 4.0, which required fixing breaking changes in the builder API. The MSRV must also be bumped to 1.60.0.


sharkdp (approved):
Thank you. This looks great.

PR #178, opened 2022-11-27
[feature] Implement alternate bases

### LINKED ISSUE #147: Decimal values in place of the hex characters
In the past few weeks, I've been working with odd formats that make use of Pascal strings in the encoding. And working with hex numbers while also having to count out distance and such is cumbersome, especially with changing values and if you don't have a website/calculator/piece of paper/another shell at hand. Normal `hexdump` supports specifying a custom format (specifying which is egregiously complex...), but thanks to [this answer on StackOverflow](https://stackoverflow.com/a/26692208), there is a pre-cooked way to gather single-byte decimal (which isn't exposed on `hexdump`'s normal interface as a shortcut, unlike single-digit octal and two-digit decimal...) dumps together with the characters:

```
$ echo "sharkdp/hexyl" | hexdump -e'"%07.8_ad  " 8/1 "%03d " "  " 8/1 "%03d " "  |"' -e'16/1  "%_p"  "|\n"' -e '"%07.8_Ad\n"'
00000000  115 104 097 114 107 100 112 047  104 101 120 121 108 010          |sharkdp/hexyl.|
00000014
```

(However, this format lacks the separator in the middle for the characters, and also, just like `hexdump`'s usual, lacks colours.)
One problem with the above format is "negative numbers". This is very like platform specific, but `char 128` (`\PAD`) is represented as `-128` in the output of `hexdump`. While this kind of ties into #104, it would be nicer if the decimal could be specified to always be positive (it also breaks the output formatting because of the sometimes-appearing `-` signs), i.e. between `0` and `255`.

```
00001056  000 000 000 000 000 005 -128 -128  -128 -128 -128 040 000 000 000 006  |...........(....|
```

So what I'd like to request is a slight formatting change, instead of columns of 2, columns of 3 digits, which each show the decimal value of the byte at hand. The rest should remain the same, together with the colouring, etc. On my system, conventional `hexdump` exposes the following shortcuts:

```
$ echo "sharkdp/hexyl" | hexdump -b
0000000 163 150 141 162 153 144 160 057 150 145 170 171 154 012
000000e

$ echo "sharkdp/hexyl" | hexdump -c
0000000   s   h   a   r   k   d   p   /   h   e   x   y   l  \n
000000e

$ echo "sharkdp/hexyl" | hexdump -d
0000000   26739   29281   25707   12144   25960   31096   02668
000000e

$ echo "sharkdp/hexyl" | hexdump -o
0000000  064163  071141  062153  027560  062550  074570  005154
000000e

$ echo "sharkdp/hexyl" | hexdump -x
0000000    6873    7261    646b    2f70    6568    7978    0a6c
000000e
```

Sadly, `-d` means **two-byte** decimal (but at least unsigned!), so a better shortcut character should be used, if #104 ever wants to be implemented in an invocation-compatible way. (`--word-size=2 --decimal --unsiged --little-endian` would be the mapping of `hexdump`'s `-d`.)

sharifhsn:
Fixes #147.

This PR provides the new `--base`/`-b` argument which specifies the base that the bytes should be printed in. The base options exposed by the standard library are binary, octal, decimal, and hexadecimal. If we were to add more options, then we would need to pull in some kind of dependency.

For ease of use, the argument accepts either a number (2, 8, 10, 16), a single character abbreviation (b, o, d, x), a three-character abbreviation (bin, oct, dec, hex), or the full base name (binary, octal, decimal, hexadecimal).

Part of this PR is adjusting the `--panels=auto` option to account for both base and byte grouping from #170. Now it is guaranteed to give the correct number of panels regardless of options.

I have also added a simple integration test.

Questions for review:
- Should more bases be allowed?
- What should the help text suggest for the argument format?
- Should there be more tests?

This will be my last PR until the next release of `hexyl`.


sharkdp:
> Should more bases be allowed?

I think the supported set is completely reasonable. I don't expect anyone to require base 5 or similar.

> What should the help text suggest for the argument format?

You're talking about the `value_name`? I think `B` is fine. *Maybe* for the values, we could support either the name (hexadecimal) or a number (16), but that's not really important for now. I'm completely fine with your proposal.

It seems pretty obvious, but maybe we could mention the default (`hexadecimal`).

> Should there be more tests?

I think we're good for now - thank you!

sharkdp (approved):
Thank you!

PR #180, opened 2022-12-05
Replace outdated dependencies

sharifhsn:
The `atty` crate is unmaintained, so I have changed it to its equivalent [`is-terminal`](https://lib.rs/crates/is-terminal), which is a simple one-liner.

The ecosystem seems to have embraced `rustix` as a replacement for interacting with `libc` directly, so I've chosen to  replace it, which is another one-liner.

This is my *actual* last PR 😅


sharkdp (approved):
Thank you!

sharkdp:
For some reason, CI didn't run on this branch (??) - and I missed that. I reverted the `rustix` commit on master, since it did not compile on Windows: https://github.com/sharkdp/hexyl/actions/runs/3618076985

Issue #89, opened 2020-05-17
Handle sparse files


clarfon:
Because hexyl truncates repeating sections, it would be nice to be able to have hexyl quickly skip over these sections instead of scanning them byte-by-byte.

sharkdp:
So this is a request for `hexyl` to be faster?

sharkdp:
Closing for now, as it's unclear to me.

clarfon:
Sorry I never got to this. And, well, yes.

All the major filesystems expose the concept of sparse files, where the file is fragmented on disk and entire sections are marked as being zeroes without actually being stored.

For example, take an 8GiB disk image that is freshly formatted from a zeroed-out disk, with one or two files written to it that are both a few KiB. The actual file will only take up a few KiB on disk because the OS simply marks the file as being all zero except for a few sections.

While hexyl going through the file naively would take several minutes of reading the zero bytes, ultimately it would truncate these sections in the actual output. Reading these files would become feasible if hexyl used the OS subroutines to query the file for its zero ranges, then used this information to skip reading those unnecessarily.

VioletDarkKitty:
A mode where it printed out "x number of skipped duplicate bytes" or something would be useful for this. I can imagine someone will still want a verbose output so it might be best to not be default?

sharifhsn:
Unfortunately, due to the type signature of the `print_all` function, this isn't really possible. In order to access sparse file search on Windows and Linux (haven't checked MacOS) the data type must be a file. However, the `print_all` function accepts any type that implements `Read`, which makes it impossible to check whether it's reading a file or not. One way to change this is to make `print_all` accept the `Input` enum, but that would limit any generic usage of `hexyl` to read any data except a `File` or `Stdin`.

sharkdp:
> While hexyl going through the file naively would take several minutes of reading the zero bytes, ultimately it would truncate these sections in the actual output. Reading these files would become feasible if hexyl used the OS subroutines to query the file for its zero ranges, then used this information to skip reading those unnecessarily.

@clarfonthey Is this really the case? Can you provide any benchmark/timing results? Ideally, in a reproducable setup.

clarfonthey:
I mean, it's pretty easy to just `truncate -s 2G test` and then run `hexyl test` to see that while the file takes up no space on disk, it will loop for quite a while.

sharkdp:
[...]

PR #185, opened 2022-12-08
experiment: Sparse file support

sharifhsn:
I decided to take a stab at #89 just to see how feasible it is. In order to facilitate this, I have temporarily changed the API of `print_all` to take `Input` with a new generic `Box <dyn Read>` variant as a fallback.

So far, this implementation is only able to use sparseness when there is non-sparse content at the end of the file, and the rest of the file, including the start, is sparse. Otherwise, it will cause strange errors.

It also only works for Linux, because this was the simplest OS to implement sparse support for.

How to test this:
```
truncate -s 16G startzero
echo "Lorem ipsum" >> startzero
cargo run --release -- startzero
```

The position will be slightly wrong but it will finish extremely quickly.

Tasks:

- [x] correct position indexing
- [x] handle fully sparse files
- [x] handle when the file starts with non-sparse data
- [x] handle non-sparse data interspersed in files (this might work already to some extent)
- [x] fully support Linux (caveat: does not work with `BufReader`)
- [ ] fully support Windows
- [ ] fully support MacOS


sharkdp:
> I decided to take a stab at #89 just to see how feasible it is. In order to facilitate this, I have temporarily changed the API of `print_all` to take `Input` with a new generic `Box <dyn Read>` variant as a fallback.

Just a few comments for now:

- I said from the beginning that I'm only going to maintain hexyl-as-a-library on a best-effort basis. I have no problems with breaking changes to the API. Especially if it can advance hexyl-as-an-application. It's also not like we have hundreds of [dependents](https://crates.io/crates/hexyl/reverse_dependencies).
- I really appreciate your effort here. If you are interested in the problem itself - great! But I just want to warn you that this might very well turn out to be too complex (or too risky in terms of bugs) for an actual merge. I'm not saying it definitely will. Just want to mention that this is a possibility. 
- Maybe it would be useful to use sth like `nix` instead of `libc` directly?

sharifhsn:
> If you are interested in the problem itself - great! But I just want to warn you that this might very well turn out to be too complex (or too risky in terms of bugs) for an actual merge.

Yeah, I'm pretty aware of this, at least at my progress so far. That's why I'm labeling this PR as an experiment, I just want to see how possible it is. At least so far, it seems to work fine on Linux (with a major caveat that I'm trying to work around at the moment).

I hadn't looked into `nix` before, only `rustix`. It seems that `nix` supports `SEEK_DATA` while `rustix` doesn't, so I think I'll replace this use of `libc` with `nix` (and also maybe raise an issue for `rustix`).

PR #189, opened 2023-04-18
Support both little and big endian dump

RinHizakura:
In #170, we solve a part of the demand for #104, which provided the option to group bytes.  In this patch, I am trying to complete the remaining requirement for endianness.

### DIFF
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -434,7 +434,9 @@ fn run() -> Result<()> {
         )
     };
 
-    let little_endian_dump = *matches.get_one::<bool>("little_endian_dump").unwrap_or(&false);
+    let little_endian_dump = *matches
+        .get_one::<bool>("little_endian_dump")
+        .unwrap_or(&false);
     let stdout = io::stdout();
     let mut stdout_lock = BufWriter::new(stdout.lock());
 
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -203,6 +203,15 @@ fn run() -> Result<()> {
                     width but still leave some space to the right.\nCannot be used with other \
                     width-setting options.",
                 ),
+        )
+        .arg(
+            Arg::new("little_endian_dump")
+                .short('e')
+                .action(ArgAction::SetTrue)
+                .help(
+                    "Print out the data in little endian format. Otherwise the data will be showed \
+                    in big endian format by default.",
+                ),
         );
 
     let matches = command.get_matches();
@@ -425,6 +434,7 @@ fn run() -> Result<()> {
         )
     };
 
+    let little_endian_dump = *matches.get_one::<bool>("little_endian_dump").unwrap_or(&false);
     let stdout = io::stdout();
     let mut stdout_lock = BufWriter::new(stdout.lock());
 
@@ -437,6 +447,7 @@ fn run() -> Result<()> {
         .num_panels(panels)
         .group_size(group_size)
         .with_base(base)
+        .little_endian_dump(little_endian_dump)
         .build();
     printer.display_offset(skip_offset + display_offset);
     printer.print_all(&mut reader).map_err(|e| anyhow!(e))?;
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -209,7 +209,7 @@ fn run() -> Result<()> {
                 .short('e')
                 .action(ArgAction::SetTrue)
                 .help(
-                    "Print out the data in little endian format. Otherwise the data will be showed \
+                    "Print out the data in little endian format. Otherwise the data will be shown \
                     in big endian format by default.",
                 ),
         );
--- a/src/bin/hexyl.rs
+++ b/src/bin/hexyl.rs
@@ -190,8 +190,17 @@ fn run() -> Result<()> {
                 .help(
                     "Whether to print out groups in little-endian or big-endian \
                      format. This option only has an effect if the '--group-size' \
-                     is larger than 1.",
+                     is larger than 1. '-e' can be used as an alias for \
+                     '--endianness=little'.",
                 ),
+        )
+        .arg(
+            Arg::new("little_endian_format")
+                .short('e')
+                .action(ArgAction::SetTrue)
+                .overrides_with("endianness")
+                .hide(true)
+                .help("An alias for '--endianness=little'."),
         )
         .arg(
             Arg::new("base")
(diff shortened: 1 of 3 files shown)


src/bin/hexyl.rs
```diff
@@ -203,6 +203,15 @@ fn run() -> Result<()> {
...
                 ),
+        )
+        .arg(
+            Arg::new("little_endian_dump")
+                .short('e')
+                .action(ArgAction::SetTrue)
+                .help(
+                    "Print out the data in little endian format. Otherwise the data will be showed \
+                    in big endian format by default.",
+                ),
```
sharkdp, line 214: Is this only usable with `--group-size`? If so, we should specify that using https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires.

Also, should we maybe introduce a `--endianness=little/big` option instead (where we can still discuss what the default should be)? We could potentially still have `-e` as an alias for `--endianness=little` if we choose `big` as the default.

src/lib.rs
```diff
@@ -159,6 +159,7 @@ pub struct PrinterBuilder<'a, Writer: Write> {
     panels: u64,
     group_size: u8,
     base: Base,
+    little_endian_dump: bool,
```
sharkdp, line 162: I would prefer if we add a `enum Endianness { Little, Big }` instead of using a bool internally.

src/lib.rs
```diff
@@ -216,6 +218,11 @@ impl<'a, Writer: Write> PrinterBuilder<'a, Writer> {
         self
     }
 
+    pub fn little_endian_dump(mut self, little_endian_dump: bool) -> Self {
```
sharkdp, line 221: Similar here, let's call this `endianness` and let it accept a `Endianness` enun argument?

src/lib.rs
```diff
@@ -523,8 +535,27 @@ impl<'a, Writer: Write> Printer<'a, Writer> {
...
+        }
+    }
+
     pub fn print_bytes(&mut self) -> io::Result<()> {
-        for (i, &b) in self.line_buf.clone().iter().enumerate() {
+        let mut buf = self.line_buf.clone();
+
+        if self.little_endian_dump {
+            // reorder the buffer to the little endian format
+            self.reorder_buf_to_le(&mut buf);
```
sharkdp, line 555: The character panel is not affected by this reordering. Is this by design? `xxd` does the same, but I'm not entirely sure why?

tests/integration_tests.rs
```diff
@@ -292,12 +292,12 @@ mod display_settings {
     }
 }
 
-mod group {
+mod group_and_endian {
```
sharkdp, line 295:
```suggestion
mod group_and_endianness {
```

sharkdp (changes requested):
Thank you very much! This looks good.

src/bin/hexyl.rs
RinHizakura: I think this option doesn't have to be depending on `--group-size`. Because if we don't specify the group size in the option, we'll default it to one. In other words, `hexyl -g 1 file` and `hexyl file` are the same.


The reason that I choose to use big-endian as default is because of the design of `xxd`:
```
$ xxd --help
...
Options:
    -e          little-endian dump (incompatible with -ps,-i,-r).
```
However, it would be fine if `hexyl` want to have a different design for the dump of endianness. Please tell me which you would prefer.

src/lib.rs
RinHizakura: In my option, the endianness only makes sense for the "grouping" bytes representation. For the character panel, each character will directly correspond to one single byte, which has no difference between big or little endianness.

src/bin/hexyl.rs
sharkdp: > I think this option doesn't have to be depending on --group-size. Because if we don't specify the group size in the option, we'll default it to one. In other words, hexyl -g 1 file and hexyl file are the same.

yes. But Endianness won't have any effect if the group size is 1, right?


> However, it would be fine if hexyl want to have a different design for the dump of endianness. Please tell me which you would prefer.

Let's follow xxd here for now, but I would still prefer a `--endianness=little/big` option instead of a `--little-endian-dump` flag.

src/lib.rs
sharkdp: Ok. I just thought it would be confusing as there is no 1:1 correspondence anymore when using little-endian mode. But let's leave things as they are for now.

src/bin/hexyl.rs
```diff
@@ -203,6 +203,15 @@ fn run() -> Result<()> {
                     width but still leave some space to the right.\nCannot be used with other \
                     width-setting options.",
                 ),
+        )
+        .arg(
+            Arg::new("little_endian_dump")
+                .short('e')
+                .action(ArgAction::SetTrue)
+                .help(
+                    "Print out the data in little endian format. Otherwise the data will be showed \
```
sharkdp, line 212:
```suggestion
                    "Print out the data in little endian format. Otherwise the data will be shown \
```

src/lib.rs
```diff
@@ -254,6 +267,8 @@ pub struct Printer<'a, Writer: Write> {
     group_size: u8,
     /// The number of digits used to write the base.
     base_digits: u8,
+    /// Whether to print the bytes in little endian
```
sharkdp, line 270:
```suggestion
    /// Whether to show groups in little or big endian ordering.
```

src/lib.rs
```diff
@@ -523,8 +540,27 @@ impl<'a, Writer: Write> Printer<'a, Writer> {
         Ok(())
     }
 
+    fn reorder_buf_to_le(&self, buf: &mut Vec<u8>) {
```
sharkdp, line 543: I'd prefer a non-abbreviated name
```suggestion
    fn reorder_buffer_to_little_endian(&self, buf: &mut Vec<u8>) {
```

src/bin/hexyl.rs
RinHizakura: > But Endianness won't have any effect if the group size is 1, right?

Sure. If we want to inform the user that this would be no effect, let's use the [requires](https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires) method.

> Let's follow xxd here for now, but I would still prefer a --endianness=little/big option instead of a --little-endian-dump flag.

I change the command line option to `--endianness=little/big` now.

sharkdp: I thought about it again and now think that your initial version was actually better (sorry). Because *not* using `requires` allows someone to set an alias like `alias hexyl="hexyl --endianness=little"`, and use that even if `-g` is not present.

Instead, we can mention that it only has an effect if group size > 1. 

I'll make the necessary adjustments.

sharkdp (approved):

Issue #194, opened 2023-06-13
Support for different code pages / character tables


sharkdp:
I found this [cool blog post](https://www.azabani.com/2020/11/15/xd.html) by @delan today and I think we should experiment with adding support for different code pages to `hexyl`. We currently print different *categories* of bytes using `⋄`, `×` or `•`, but it would be nice to be able to distinguish each byte. Maybe we could add a `--character-table` command line option (or similar) that could be used to switch to different byte => Unicode code point mappings. The default would (probably?) be what we have right now, but we could then set e.g. `--character-table codepage-437` to use [Code page 437](https://en.wikipedia.org/wiki/Code_page_437).

PR #195, opened 2023-06-13
Implement character table control and codepage 437 option

sharifhsn:
Implements #194

Additional comments:
- I've directly copy-pasted the codepage table from the repository in the blog post you linked. I'm not sure how to properly give credit for it.
- I replaced the null character with the ⋄ character since null doesn't appear in printed output, which is pretty common.


sharifhsn:
Here's a visual image of the difference.
image

sharkdp:
Wow! That was fast :smile: 

Thank you so much.

> I've directly copy-pasted the codepage table from the repository in the blog post you linked. I'm not sure how to properly give credit for it.

I think I saw a MIT license in the repository earlier today? If so, I would just copy the license text verbatim and paste it into a comment preceding the table. And add a reference to the original source file + author, if not present in the license text. Maybe let's also describe the manual modifications we did in that comment.

src/bin/hexyl.rs
```diff
@@ -209,6 +209,17 @@ fn run() -> Result<()> {
...
                 .help("An alias for '--endianness=little'."),
         )
+        .arg(
+            Arg::new("character-table")
+                .long("character-table")
+                .value_name("FORMAT")
+                .value_parser(["codepage-437", "ascii-only"])
+                .help(
+                    "The character table that should be used. 'ascii-only' \
+                    will show dots for non-ASCII characters, and 'codepage-437 \
```
sharkdp, line 219:
```suggestion
                    will show dots for non-ASCII characters, and 'codepage-437' \
```

src/bin/hexyl.rs
```diff
@@ -209,6 +209,17 @@ fn run() -> Result<()> {
                 .hide(true)
                 .help("An alias for '--endianness=little'."),
         )
+        .arg(
+            Arg::new("character-table")
+                .long("character-table")
+                .value_name("FORMAT")
+                .value_parser(["codepage-437", "ascii-only"])
```
sharkdp, line 216: I think we can define a default here? This way, we don't have to use `.unwrap_or(&String::from("ascii-only"))` below

src/lib.rs
```diff
@@ -74,17 +103,27 @@ impl Byte {
...
+            CharTable::CP437 => CP437[self.0.to_ne_bytes()[0] as usize],
         }
+        // match self.category() {
+        //     Null => '⋄',
+        //     AsciiPrintable => self.0 as char,
+        //     AsciiWhitespace if self.0 == 0x20 => ' ',
+        //     AsciiWhitespace => '_',
+        //     AsciiOther => '•',
+        //     NonAscii => '×',
+        // }
```
sharkdp, line 126: remove?

src/lib.rs
```diff
@@ -182,6 +222,7 @@ impl<'a, Writer: Write> PrinterBuilder<'a, Writer> {
             group_size: 1,
             base: Base::Hexadecimal,
             endianness: Endianness::Big,
+            char_table: CharTable::AsciiOnly,
```
sharkdp, line 225: I think I'd prefer this to be spelled out: character_table and CharacterTable.

src/lib.rs
```diff
@@ -30,6 +53,12 @@ pub enum ByteCategory {
     NonAscii,
 }
 
+#[derive(Copy, Clone)]
+pub enum CharTable {
+    AsciiOnly,
+    CP437,
+}
```
sharkdp, line 60: Since this is part of the public API, let's mark it [`non_exhaustive`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) such that we can add more variants in the future without introducing breaking changes (technically).

delan:
Nice! I’m glad this idea is catching on.

Note that ␀ is already a printable character, since it’s U+2400 SYMBOL FOR NULL rather than U+0000 NULL. In fact, the whole range U+2400 through U+241F are printable representations of the C0 controls U+0000 through U+001F.

sharkdp:
> Nice! I’m glad this idea is catching on.

Thank you very much for commenting here and for the original idea! A while ago, I built [binocle](https://github.com/sharkdp/binocle), which uses a similar idea in a graphical application (distinctly colored pixels for each byte), but I never thought about translating this idea to the character panel of `hexyl`.

I'm now thinking that it might also be cool to have a mode in `hexyl` where we have distinct colors for *every* byte (not just for different categories). We could then print █ (U+2588) for every byte in the character table, but colorize it differently using ANSI escape sequences. As for this PR, we could maybe already add a `--character-table block` (or similar) variant where we simply print █ for each byte. For the GIF example above, it would look like this:

image

> Note that ␀ is already a printable character, since it’s U+2400 SYMBOL FOR NULL rather than U+0000 NULL. In fact, the whole range U+2400 through U+241F are printable representations of the C0 controls U+0000 through U+001F, though I chose to use the cp437 dingbats for the others.

We use (some of) those U+2400 … U+241F symbols in `bat --show-all`. I like the idea. But at least with my terminal emulator / font, they are usually really hard to read. Even at large font sizes, I can barely make out the "NUL" characters. This is why we chose `⋄` for `hexyl`. @sharifhsn but we should update the comment for consistency: "modified to use the '⋄' character instead of '␀'"

> FFh is also visually indistinguishable from 20h in this table, so for the [Rewind path interpreter](https://bucket.daz.cat/rewind/path.html#display), we used U+FB00 LATIN SMALL LIGATURE FF (ﬀ) for FFh.

@sharifhsn I think we should integrate the same adaptation here, if you agree?

sharifhsn:
> @sharifhsn but we should update the comment for consistency: "modified to use the '⋄' character instead of '␀'"

Makes sense.

> @sharifhsn I think we should integrate the same adaptation here, if you agree?

Ditto.

> I'm now thinking that it might also be cool to have a mode in hexyl where we have distinct colors for every byte (not just for different categories). We could then print █ (U+2588) for every byte in the character table, but colorize it differently using ANSI escape sequences. As for this PR, we could maybe already add a --character-table block (or similar) variant where we simply print █ for each byte.

Great idea! I've implemented this using the `xterm` colors as there are 256 distinct colors, which maps perfectly to the number of distinct bytes. I'm open to a change in color scheme. Here's how it looks now:

image

sharkdp:
This is so cool.

Especially when combined with `--border none`, `--no-squeezing`, potentially `--group-size 8` and `--panels 8` on a wide screen:

image

But I would appreciated if we could focus this PR on `--character-table` and finish that first. Add a few basic tests. Make sure that it plays nicely with other features like e.g. squeezing. For example, I noticed that the squeezing character gets replaced by a block when `--character-table=block`. And maybe we should quickly check if this doesn't introduce any performance regression for the default case.

We can then explore the colorization part in a separate PR.

sharifhsn:
I added a test for code page 437, and I've fixed the issue with the asterisk, which was a leftover bug from my original implementation anyway 😅

I've also refactored the color components into a new module, since they're taking up a lot of visual space and have separated concerns anyway.

With regards to performance, preliminary benchmarks show that performance is within noise for the default case and for code page 437, and significantly slower with block coloring (due to the fact that the color changes every single byte).

|:---|---:|---:|---:|---:|
| `target/release/hexyl --character-table codepage-437 data` | 2.246 ± 0.087 | 2.169 | 2.442 | 1.00 |
| `target/release/hexyl --character-table block data` | 4.412 ± 0.120 | 4.229 | 4.567 | 1.96 ± 0.09 |
| `target/release/hexyl data` | 2.490 ± 0.357 | 2.195 | 3.234 | 1.11 ± 0.16 |
| `hexyl data` | 2.274 ± 0.142 | 2.071 | 2.579 | 1.01 ± 0.07 |

With regards to colorization and the block table, it's all the same machinery. I would prefer to push this all in one PR unless there's some outstanding issue with the block table that can't be resolved quickly.

sharkdp:
> I've also refactored the color components into a new module, since they're taking up a lot of visual space and have separated concerns anyway.

Thank you for the updates. I would still prefer to move the colorization part to a new PR.

sharifhsn:
@sharkdp The colorization of characters has been factored out. This PR should be ready to merge.

sharkdp (commented):
Thank you very much! Just two minor questions.

PR #197, opened 2023-06-20
Correctly handle discontinuous input i.e. stdin

### LINKED ISSUE #196: Incorrect output when reading from terminal
Hexyl is giving incorrect output in some cases when reading terminal input. I've seen it when the input has multiple newlines at the end. The input of `Test[Enter][Enter][Enter][Enter][CTRL-D]` is one such case:

```
❯ hexyl
Test



│00000000│ 54 65 73 74 0a 0a 00 00 ┊ 00 00 00 00 00 00 00 00 │Test__⋄⋄┊⋄⋄⋄⋄⋄⋄⋄⋄│
│00000010│ 0a 0a 73 74 0a 0a 00 00 ┊ 00 00 00 00 00 00 00 00 │__st__⋄⋄┊⋄⋄⋄⋄⋄⋄⋄⋄│
```

I've been able to track it back to 15e8abfc9afe95a221c9c9a265cf3fb58fed0470 through 5f82942f98fc69af8f76701715053aee2451d62c, with it being introduced in that set of changes.

sharifhsn:
Fixes #196.

This PR properly handles the case when the reader is still available after returning an incomplete amount of data. When I initially rewrote the architecture of `hexyl`, I only handled spurious incomplete reads by adding a second check. However, `stdin` is a reader that can return incomplete reads multiple times.

This PR changes the second check to a loop that repeatedly reads until either the line buffer is filled or the reader is exhausted.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -585,7 +585,7 @@ impl<'a, Writer: Write> Printer<'a, Writer> {
                     }
                     let mut leftover = n;
                     // loop until input is ceased
-                    break loop {
+                    if let Some(s) = loop {
                         if let Ok(n) = buf.read(&mut self.line_buf[leftover..]) {
                             leftover += n;
                             // there is no more input being read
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much.

PR #201, opened 2023-08-21
Move tests to a separate `#[cfg(test)]` module

GrigorenkoPV:
So that the compiler doesn't have to go over their code during `cargo check` or `cargo build` or `cargo run`.


sharkdp:
Thank you

PR #202, opened 2023-08-21
Vet dependencies

GrigorenkoPV:
Two commits.

1. Deduplicate `predicates`

hexyl depends on `predicates v2.0`.
hexyl depends on `assert_cmd v2.0`, which depends on `predicates v3.0`.
So we end up compiling two versions of the same crate.

Fixed by bumping the direct dependency from `2.0` to `3.0`. Shouldn't have broken anything, from what I can tell after reading [the changelog](https://github.com/assert-rs/predicates-rs/blob/v3.0.3/CHANGELOG.md) and seeing how little the library is actually used in this codebase.

2. Run `cargo update`

Not necessary, but since we're at it...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,7 +75,7 @@ dependencies = [
  "anstyle",
  "bstr",
  "doc-comment",
- "predicates 3.0.3",
+ "predicates",
  "predicates-core",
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you!

PR #211, opened 2023-12-11
`/dev/zero` fix (final)

### LINKED ISSUE #200: `hexyl /dev/zero` hangs
So does
```shell
hexyl < /dev/zero
```
or
```shell
cat /dev/zero | hexyl
```

I guess it has to do with hexyl's feature where it skips null rows and prints only the first and the last.

sharifhsn:
Fixes #200.

Instead of flushing on every single print, it flushes only when the first line is printed. This produces the expected behavior for `hexyl /dev/zero`:

```
❯ target/debug/hexyl /dev/zero
│00000000│ 00 00 00 00 00 00 00 00 ┊ 00 00 00 00 00 00 00 00 │⋄⋄⋄⋄⋄⋄⋄⋄┊⋄⋄⋄⋄⋄⋄⋄⋄│
^C
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -621,7 +621,6 @@ impl<'a, Writer: Write> Printer<'a, Writer> {
             }
             if is_empty {
                 self.print_header()?;
-                is_empty = false;
             }
 
             // squeeze is active, check if the line is the same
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you!

PR #213, opened 2023-12-12
Squeeze identical lines

### LINKED ISSUE #186: Identical lines not squeezed
The help indicates:

> --no-squeezing Displays all input data. Otherwise any number of groups of output lines which would be identical to the preceding group of lines, are replaced with a line comprised of a single asterisk.

It seems that lines are squeezed when they all are filled with the same single byte (for example, all 0x00 or all 0xFF). However, lines which are not filled with a single repeated byte can still be identical to previous lines. In these cases they do not appear to be squeezed.

I'll attach an example file, which contains rows filled with the two-byte word 0x1400 starting at position 00000020. If the standard two panels are displayed, I would expect this file to have data squeezed until position 00000fa0.

[example.tar.gz](https://github.com/sharkdp/hexyl/files/10431306/example.tar.gz)

sharifhsn:
Fixes #186.

This brings the squeezing feature to parity with `hexdump`. Lines that contain identical data should be squeezed, not just when they are all filled with a single byte.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,7 +36,7 @@ pub enum Endianness {
     Big,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(PartialEq)]
 enum Squeezer {
     Print,
     Delete,
@@ -657,7 +657,6 @@ impl<'a, Writer: Write> Printer<'a, Writer> {
                         continue;
                     }
                 } else {
-                    write!(self.writer, "self")?;
                     self.squeezer = Squeezer::Ignore;
                 }
             }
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

Some tests are currently failing..

sharifhsn:
Looks like the issues are on 32-bit systems. I'll investigate.

sharkdp:
Note: there is also #218

GrigorenkoPV (changes requested):

sharifhsn:
Closing as I'm no longer actively contributing to hexyl.

Issue #155, opened 2022-04-30
Shell completions


Freed-Wu:
Any shell completion is good.

Thanks!

Same issues:

- https://github.com/sharkdp/vivid/issues/86

sharkdp:
Sounds good. Could be easily autogenerated using https://docs.rs/clap_complete/latest/clap_complete/

Freed-Wu:
Fixed by #251

PR #214, opened 2024-01-22
feat: Generate shell completions on build

FilipRazek:
My attempt at closing #155


sharkdp:
Thank you very much!

It would be great if we could also integrate the generated shell completion files in the artifacts we built when we make a release. You can find examples how we do this in some other projects of mine, like hyperfine (see [https://github.com/sharkdp/hyperfine a](https://github.com/sharkdp/hyperfine/blob/master/.github/workflows/CICD.yml) and search for complete/completion)

FilipRazek:
Thanks for the hint, I tried to update the `CICD` file, tell me if this would work for you

sharkdp:
You can find the generated artifacts here at the bottom: https://github.com/sharkdp/hexyl/actions/runs/8010537770

FilipeBento:
Hello! Will this go through? 😄 (I don't have enough knowledge to implement this on my own, I'm sorry)

sharkdp:
@FilipRazek Sorry for not coming back to this. Would it be possible to update this to the latest state of master? I'd love to integrate this.

FilipRazek:
I'm sorry, I don't know how to integrate clap_complete with the new API. I will look into it but it doesn't seem to be a straightforward change

sharkdp:
I'm closing this due to inactivity. Please feel free to comment in case it should be re-opened.

PR #226, opened 2024-08-04
Add codepage 1047 for EBCDIC

v1gnesh:
Thanks for yet another excellent tool.

**EDIT:** How much ~would~ work is it to make the bytes show up bolder, like with xxd?


sharkdp:
Thank you for your contribution. What is the advantage of that code page over the existing formats?

> How much would is it to make the bytes show up bolder, like with xxd?

what exactly do you mean?

v1gnesh:
> What is the advantage of that code page over the existing formats?

What existing formats do you mean?
The advantage is that I can now see EBCDIC strings with hexyl, as is possible with xxd.
If you mean why this particular EBCDIC character set (when there are many EBCDIC character sets), this (1047) is the default on a widely used (but not known) platform that uses EBCDIC (mainframe).

Updated the first comment; I meant 'work' instead of 'would'.

`alias xxd='xxd -d -E -o 1 -g 1'`

<summary>with xxd (see that it's bold)</summary>
<details>

image
</details>

`alias hxl='$HOME/_dev/hexyl/target/release/hexyl --character-table=codepage-1047 --panels 4'`

<summary>with this patch in hexyl</summary>
<details>

image

</details>

sharkdp:
> What existing formats do you mean?
> The advantage is that I can now see EBCDIC strings with hexyl, as is possible with xxd.

Ah, I didn't notice that the alphanumeric block somewhere else compared to the ASCII encoding. Thanks.

> with xxd (see that it's bold)

Ah. You seem to have a different version of xxd. For me, it's not bold. I don't think it would be a lot of work. Can we please open a new ticket for this?

v1gnesh:
@sharkdp Congrats on you joining Astral. Very, very satisfying that meticulous & hard work is matched with worthy reward & recognition :muscle: 

Would you mind publishing a release, please, with this addition in?

sharkdp:
> @sharkdp Congrats on you joining Astral. Very, very satisfying that meticulous & hard work is matched with worthy reward & recognition 💪

Thank you!



> Would you mind publishing a release, please, with this addition in?

Done.

PR #234, opened 2024-11-20
fix: throw an error when try to view a directory

Integral-Tech:


Integral-Tech:
@sharkdp

src/main.rs
```diff
@@ -234,7 +234,14 @@ fn run() -> Result<()> {
     let stdin = io::stdin();
 
     let mut reader = match opt.file {
-        Some(filename) => Input::File(File::open(filename)?),
+        Some(filename) => {
+            let file = File::open(&filename)?;
+            if file.metadata()?.is_dir() {
+                return Err(anyhow!("{} is a directory.", filename.display()));
+            }
```
sharkdp, line 241: Some minor comments:
* we can check `is_dir` on the path itself, instead of retrieving it from the `File`
* we can use `bail!(…)` instead of `return Err(anyhow!(…))`

```suggestion
            if filename.is_dir() {
                bail!("'{}' is a directory.", filename.to_string_lossy());
            }
            let file = File::open(&filename)?;
```

sharkdp:
Thank you!

PR #235, opened 2024-12-13
PR #229 • Add option to print color table

### LINKED ISSUE #229: Add option to print the color table
It would be nice to have an option like `--print-color-table` that, well, prints the color table 🙂 It would mean not having to look it up in the [readme](https://github.com/sharkdp/hexyl?tab=readme-ov-file#color-reference) (which also seems to be missing what the color red denotes in the output).

Another benefit would be that it would show the color coding on the current system, which might be different from what's shown in the readme.

sahinfalcon:
This PR adds the option to print a colour table, fixes #229

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,7 +16,10 @@ use terminal_size::terminal_size;
 
 use hexyl::{Base, BorderStyle, CharacterTable, Endianness, Input, PrinterBuilder};
 
-use hexyl::{COLOR_NULL, COLOR_RESET, COLOR_ASCII_PRINTABLE, COLOR_ASCII_WHITESPACE, COLOR_ASCII_OTHER, COLOR_NONASCII};
+use hexyl::{
+    COLOR_ASCII_OTHER, COLOR_ASCII_PRINTABLE, COLOR_ASCII_WHITESPACE, COLOR_NONASCII, COLOR_NULL,
+    COLOR_RESET,
+};
 
 #[cfg(test)]
 mod tests;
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -468,6 +481,43 @@ impl From<NonNegativeI64> for u64 {
     }
 }
 
+fn print_color_table() -> io::Result<()> {
+    let stdout = io::stdout();
+    let mut stdout_lock = BufWriter::new(stdout.lock());
```
sharkdp, line 486: I know we use this pattern already, but this could be
```rs
let mut out = BufWriter::new(io::stdout().lock());
```
in recent versions of Rust.

src/main.rs
```diff
@@ -179,6 +184,10 @@ struct Opt {
         conflicts_with("panels")
     )]
     terminal_width: Option<NonZeroU64>,
+
+    /// Print a table showing how different types of bytes are colored.
+    #[arg(short('t'), long)]
```
sharkdp, line 189: Can we remove the short flag here, in case we want to use it for something that is used more frequently in the future?
```suggestion
    #[arg(long)]
```

sharkdp (changes requested):
Thank you very much!

sharkdp (approved):

thernstig:
@sharkdp @sahinfalcon I am using the default colors (a default install) and see this:

image

Is there a reason whitespace and control characters does not default to two different colors?

sharkdp:
> Is there a reason whitespace and control characters does not default to two different colors?

No particular reason. I guess I didn't see the need to distinguish these categories so far, and apparently noone else did either.

thernstig:
> > Is there a reason whitespace and control characters does not default to two different colors?
> 
> No particular reason. I guess I didn't see the need to distinguish these categories so far, and apparently noone else did either.

Ok, I though that since they had different icons, also color would be a way to distinguish them. Why I think so I do not even know? Maybe something with cognitive association or some other psychological reason 😆

Issue #244, opened 2025-04-05
memory allocation of 4340410370284600376 bytes failed


Jiehong:
Hello,

A typo lead to a bug discovery:

```sh
hexyl --terminal-width 2 1.txt
memory allocation of 4340410370284600376 bytes failed
```

It seems any terminal width less than 10 leads to a memory allocation issue.

sharkdp:
Thank you for reporting this

selfup:
How large is the `1.txt` file? I think this has to do with the `(terminal_width - offset)` calculation taking a quick peek at main.

sharkdp:
You can reproduce it with
```
▶ hexyl --terminal-width 2 <(echo "test")
memory allocation of 4340410370284600376 bytes failed
```

Looks very much just like an underflow in that subtraction, yes.

thomwiggers:
is this not resolved now?

PR #245, opened 2025-04-08
Check if terminal_width is less than offset and return 1 (see #244)

selfup:
In issue #244 a bug of what seems to be an underflow has been found when the terminal-width is quite small.

```console
--terminal-width 2 <(echo "test")
memory allocation of 4340410370284600376 bytes failed
```

I added a simple check in `max_panels_fn`.

Here is the fix working:

```console
$ cargo run -- --terminal-width 2 <(echo "test")
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/hexyl --terminal-width 2 /dev/fd/11`
┌────────┬─────────────────────────┬────────┐
│00000000│ 74 65 73 74 0a          │test_   │
└────────┴─────────────────────────┴────────┘
```

image

---

`cargo test` still passes (omitted noise):

```
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 38 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.19s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Time to see if CI passes as well.

Let me know if there are any other checks we'd like to do outside of existing tests/CI.

Thanks!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -362,12 +362,7 @@ fn run() -> Result<()> {
         } else {
             ((8 / group_size) * (base_digits * group_size + 1)) + 2
         };
-
-        if terminal_width < offset {
-            return 1;
-        }
-
-        if (terminal_width - offset) / col_width < 1 {
+        if (terminal_width.saturating_sub(offset)) / col_width < 1 {
             1
         } else {
             (terminal_width - offset) / col_width
--- a/src/main.rs
+++ b/src/main.rs
@@ -362,7 +362,10 @@ fn run() -> Result<()> {
         } else {
             ((8 / group_size) * (base_digits * group_size + 1)) + 2
         };
-        if (terminal_width - offset) / col_width < 1 {
+
+        if terminal_width < offset {
+            1
+        } else if (terminal_width - offset) / col_width < 1 {
             1
         } else {
             (terminal_width - offset) / col_width
--- a/src/main.rs
+++ b/src/main.rs
@@ -364,8 +364,10 @@ fn run() -> Result<()> {
         };
 
         if terminal_width < offset {
-            1
-        } else if (terminal_width - offset) / col_width < 1 {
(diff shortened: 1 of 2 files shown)


selfup:
Interesting. Clippy isn't super happy here: https://github.com/sharkdp/hexyl/actions/runs/14325281263/job/40149553728?pr=245

image

https://rust-lang.github.io/rust-clippy/master/index.html#if_same_then_else

selfup:
Clippy is now happy. Awesome

src/main.rs
```diff
@@ -362,6 +362,11 @@ fn run() -> Result<()> {
         } else {
             ((8 / group_size) * (base_digits * group_size + 1)) + 2
         };
+
+        if terminal_width < offset {
+            return 1;
+        }
```
sharkdp, line 368: It looks like you could also remove this branch and use `terminal_width.saturating_sub(offset) / col_width < 1` below?

selfup: Looks like it's only for nightly: https://doc.rust-lang.org/std/intrinsics/fn.saturating_sub.html

Oh it's available for u32 and u64 in stable:

https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_sub
https://doc.rust-lang.org/std/primitive.u64.html#method.saturating_sub

Let check it out. Thanks!

selfup: Ok, that worked great

image

`cargo test` works as well.

Time to see what CI does

sharkdp (approved):
Thank you

Issue #242, opened 2025-02-20
Add an option to output the result in C include file style


sorairolake:
I think it would be useful to have an option for `hexyl` that has the same purpose as the `-i`/`-include` option for the `xxd` command:

```sh
$ echo -n "Hello, world!" | xxd -i
  0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x2c, 0x20, 0x77, 0x6f, 0x72, 0x6c, 0x64,
  0x21
```

This output can easily be embedded as an array in many programming languages:

```rust
fn main() {
    let arr = [
        0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x2c, 0x20, 0x77, 0x6f, 0x72, 0x6c, 0x64, 0x21,
    ];
    let s = std::str::from_utf8(&arr).unwrap();
    assert_eq!(s, "Hello, world!");
}
```

sharkdp:
Sounds like a reasonable request, thank you.

PR #246, opened 2025-04-10
Add option to output result in C include file style (#242)

wpcwzy:
This pull request introduces a new feature that adds an option to output the result in C include file style, similar to the `-i` or `--include` option of the `xxd` command. This functionality allows users to generate hex dumps in a format that can be easily embedded as arrays in various programming languages.

When `hexyl` accepts input from stdin, users can expect the following output format:
```bash
$ echo -n "Hello, world!" | hexyl -i
  0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x2c, 0x20, 0x77, 0x6f, 0x72, 0x6c, 0x64,
  0x21
```

When `hexyl` accepts a filename as an argument and reads from the file, users can expect the following output format:
```bash
$ hexyl -i hello.txt
unsigned char hello_txt[] = {
  0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x2c, 0x20, 0x77, 0x6f, 0x72, 0x6c, 0x64,
  0x21
};
unsigned int hello_txt_len = 13;
```

The above behavior is consistent with the behavior of `xxd -i`.

I also added the functionality to accept input from a slice and output the bytes in C include file style, as I noticed that there is a corresponding use case in `examples/simple.rs`. When `hexyl` accepts input from a slice, its behavior is consistent with accepting input from stdin.

This is my first time contributing a new feature to an open-source project, so if there are any areas where I may have fallen short or could improve, I would really appreciate your feedback. Thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+- Add option to output result in C include file style, see #242 (@wpcwzy)
+
 ## Bugfixes
 
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -642,55 +642,26 @@ impl<'a, Writer: Write> Printer<'a, Writer> {
             // Input from a file
             // Output like `unsigned char <filename>[] = { ... }; unsigned int <filename>_len = ...;`
             IncludeMode::File(filename) => {
-                let mut data = Vec::new();
-                buf.read_to_end(&mut data)?;
                 // convert non-alphanumeric characters to '_'
                 let var_name = filename
                     .chars()
                     .map(|c| if c.is_alphanumeric() { c } else { '_' })
                     .collect::<String>();
+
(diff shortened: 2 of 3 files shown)


sharkdp:
This looks great, thank you very much! Sorry for taking such a long time to review this.

@sharifhsn Only if you are interested: do you have any thoughts on this PR?

sharifhsn:
This is nice. It's quite separate from most of `hexyl`'s functionality, so there's not much to comment on. Squeezing is inappropriate in this context and color schemes are also not relevant.

LGTM

wpcwzy:
Thank you for the review! If there's anything else I need to address or improve, please let me know. Happy to help!

sharkdp:
Sorry for the late reply. Let's merge this as-is, and add the `--output-code` option (or whatever we choose) later.

PR #247, opened 2025-04-21
Add new braille character table and color schemes

aticu:
When looking at hexdumps, I frequently find myself wishing for easier ways to distinguish between the different bytes that are just specified as `.` in a normal hexdump. I previously experimented with using the Unicode Braille characters for that. When I tried hexyl, I was immediately captured by the idea to add color and thought that color gradients would also add ways to distinguish between different bytes. I thought an image would say more than a thousand words, so I just went ahead and implemented this instead of opening an issue first. I hope this is fine for you.

Without further ado, here is what all bytes would look like with the `--character-table braille --color-scheme gradient` options after this PR is merged:

all bytes hexdump

I made several design decisions when implementing this:
1. Setting the colors and the character set are two separate command line options. (Though they go well together, so I will not distinguish specifically between them here. Which option does what is hopefully obvious.)
2.  While different bytes should have different colors, there should still be different classes of bytes, so there are several gradients used.
3. The colors should roughly get "hotter" as the byte values increase.
4. `0x7f` should not be in the same class as the other ASCII-non-printable characters as it has a very different numerical value. Picking this color was difficult, since it had to somehow fit into the general region but still be different enough so as not to be easily confusable.
5. ` ` likely makes more sense in the category of printable ASCII than non-printable ASCII, since it's very likely to be frequently contained in normal text.
6. `0x00`, `\t`, `\n` and `\r` are not really printable charaters, but still important enough to have their own symbols.
7. All other non-printable character are represented using Braille characters corresponding to their bits.

Here are some further examples to hopefully showcase how this can help pattern recognition:

UTF8:
UTF8 hexdump

UTF16LE with BOM:
UTF16LE with BOM hexdump

Randomness:
randomness hexdump

$MFT entry:
$MFT entry hexdump

hexyl binary (start):
hexyl binary hexdump

Unresolved question: How should this interact with the `--print-color-table` option? Currently this option ignores the color scheme. One option would be to show a smaller variant of the gradients with just 3-4 colors there.

Let me if you think anything should be adjusted.

EDIT: Added different character for tabs and changed 0x7f color.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Features
 
+- Add `--color-scheme` option, see #247 (@aticu)
+- Add `braille` character table, see #247 (@aticu)
+
 ## Bugfixes
 
 
--- a/src/colors.rs
+++ b/src/colors.rs
@@ -10,7 +10,7 @@ pub const COLOR_RESET: &[u8] = colors::Default::ANSI_FG.as_bytes();
 
 pub const COLOR_NULL_RGB: &[u8] = &rgb_bytes(100, 100, 100);
 
-pub const COLOR_DEL: &[u8] = &rgb_bytes(0, 255, 255);
+pub const COLOR_DEL: &[u8] = &rgb_bytes(64, 128, 0);
 
 pub const COLOR_GRADIENT_NONASCII: [[u8; 19]; 128] =
     generate_color_gradient(&[(255, 0, 0, 0.0), (255, 255, 0, 0.66), (255, 255, 255, 1.0)]);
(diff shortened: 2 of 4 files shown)


aticu:
The test failure seems unrelated? If not I am unsure how to proceed here and would be grateful for pointers.

sharkdp:
This looks cool! @sharifhsn do you have any thoughts on this?

sharifhsn:
I love the Braille idea, it's a great addition to our existing character tables.

I think making color schemes configurable is also good. I don't love the ergonomics of how this works with `Byte` and `ByteCategory`, though. These enums were never meant to accommodate alternative color schemes.

Unfortunately, I don't have the bandwidth to consider an alternative design for `hexyl` which works with this. I think this is fine to be merged for now, but if someone else wants to contribute another color scheme, these issues should be faced head on.

The good thing is that alternative color schemes are generally covered by #248 for just changing particular colors.

All that said, LGTM.

qtfkwk:
Just found this today with the 0.17.0 release (to see what changed).... this is very cool... and excellent work, @aticu!

I also like the Braille idea... that's why I made the [`bbd`](https://crates.io/crates/bbd) and [`bbd-lib`](https://crates.io/crates/bbd) crates, which look pretty basic now compared to this, but my goal was to format all bytes as Braille Unicode characters.

Thanks!

PR #249, opened 2025-05-13
Fix CI

dan-hipschman:


sharkdp (approved):
Thank you very much, especially for trying to fix the windows-gnu build and for the detailed writeup. It's fine to disable it, we still have the MSVC build.

Issue #223, opened 2024-06-12
Completions for fish shell


carneeki:
I've written a completions file that can be dropped into `~/.config/fish/completions/hexyl.fish`, or `/usr/share/fish/completions` or `/usr/share/fish/vendor_completions.d` if @sharkdp wants to include it with hexyl.

```fish
complete -c hexyl -s n -s c -l length -l bytes -f -r -d 'Only read N bytes from the input'
complete -c hexyl -s s -l skip -f -d 'Skip the first N bytes of the input'
complete -c hexyl -l block-size -f -d 'Sets the size of the block unit to SIZE'
complete -c hexyl -s v -l no-squeezing -d 'Displays all input data'
complete -c hexyl -l color -f -r -a 'always auto never force' -d 'When to use colours'
complete -c hexyl -l border -f -r -a 'unicode ascii none' -d 'Whether to draw a border with unicode, ascii or none at all'
complete -c hexyl -s p -l plain -f -d 'Display output with --no-characters, --no-position, --border=none, and --color=never'
complete -c hexyl -s C -l characters -d 'Show the character panel on the right. This is the default'
complete -c hexyl -l character-table -f -a 'default ascii codepage-437' -d 'Choose character table encoding, default is ascii'
complete -c hexyl -s P -l position -f -d 'Whether to display the position panel on the left'
complete -c hexyl -s o -l display-offset -f -r -d 'Add N bytes to the displayed file position'
complete -c hexyl -l panels -f -r -d 'Set the number of hex data panels to be displayed'
complete -c hexyl -s g -l group-size -f -r -a '1 2 4 8' -d 'Number of bytes/octets that should be grouped together'
complete -c hexyl -l endianness -f -r -a 'big little' -d 'Whether to print out groups in little-endian or big-endian format'
complete -c hexyl -s b -l base -f -r -a 'binary octal decimal hexadecimal' -d 'Sets the base, default is hexadecimal'
complete -c hexyl -l terminal-width -f -r -d 'Sets the number of terminal columns to be displayed'
complete -c hexyl -s h -l help -d 'Print help'
complete -c hexyl -s v -l version -d 'Print version'
```

sharkdp:
Thank you. What about #155/#214?

thernstig:
Seems like #155 is the way to go, would be great if you gave that a shot @carneeki 🙂

carneeki:
[...]

PR #251, opened 2025-08-03
Add shell completion

friedz:
I added a command line argument to run clap_generate for all the shells it supports out of the box. See #155 and fish being one of the shells probably also #223 

I tried to have the generation be run at compile time but while it doesn't look like it's impossible it seems impractical. Perhaps a similar result can be archived by running `hexyl --shell-completion <SHELL> > <completion file>` in the CI for all supported shells.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # unreleased 
 
 - Fix memory allocation bug when terminal width is less than 10, see #244 (@selfup)
+- Add command line argument to generate shell completion, see #155 (@friedz)
 
 ## Features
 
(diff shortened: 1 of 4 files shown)


gforien:
Hi, I frequently use hexyl and it would be nice to have shell completions
I've opened a draft PR to generate shell completions at install time for homebrew

Rondom:
Running it at compilation time would also not work when cross-compiling.

friedz:
> Running it at compilation time would also not work when cross-compiling.
I'm pretty sure it could but that would require having access to all the code necessary for the completion in the `build.rs`

Issue #262, opened 2026-03-10
Add a character table file option to enable user-provided character table(s)


qtfkwk:
Usage: `--character-table-file FILE_PATH`

Where `FILE_PATH` is a user-provided file with 256 glyphs to be used as the character table.

This allows users to create/use their own character tables without requiring support for hardcoded tables.

To improve flexibility, I'd suggest making the file JSON (or perhaps JSON5 to enable comments) and should be an array/list with 256 strings. Each string could be more than one byte to enable Unicode glyphs but should be just 1 "character".

Also add support for an environment variable that sets the character table file path.

Current character tables could be refactored to this format internally, although it's a relatively non-trivial change since they currently use an enum and code to generate characters as needed.

PR #268, opened 2026-04-10
New Character Table:  Unicode Control Pictures

asibahi:
This uses the Unicode Control Pictures block (0x24xx) for ASCII controls. For non-ASCII bytes it just reuses the same `x` icon from the default mapping. If #262 is accepted this is perhaps pointless. Note that these characters arent in all terminal fonts, but do exist in JetBrains Mono and Cascadia Code.

Using the default hexyl color scheme, this is what they look like in Cascadia Code with and without a stylistic alternative. This is a wasm module I have been looking at.

Here is Cascadia Code 

image

Here is Cascadia Code with `ss20` font feature

image

(I hope I am not being overbearing with two issues and this PR about three different topics,).


asibahi:
Just noticed https://github.com/sharkdp/hexyl/pull/250, which is pretty much the same implementation . I am unsure what the discussion in https://github.com/sharkdp/hexyl/pull/247 has to do with it as it is not a color option.

Also https://github.com/sharkdp/hexyl/pull/261 which makes a different choice (that I am not a fan of) for thee non-ascii bytes.

PR #290, opened 2026-08-15
Prefix an include-mode identifier that starts with a digit

VXNCXNX:
## What's broken

`-i` on a file whose name starts with a digit emits C that does not compile, because a C identifier cannot begin with one:

```
$ hexyl -i 9lives.bin
unsigned char 9lives_bin[] = {
...
unsigned int 9lives_bin_len = 16;

$ hexyl -i 9lives.bin > z.c && gcc -c z.c
z.c:1:15: error: invalid suffix "lives_bin" on integer constant
    1 | unsigned char 9lives_bin[] = {
      |               ^~~~~~~~~~
```

`xxd -i`, which #242 asked for parity with, handles it:

```
$ xxd -i 9lives.bin | head -1
unsigned char __9lives_bin[] = {
```

## The cause

The sanitizer in `src/lib.rs` maps non-alphanumerics to `_` but never guards the first character, so a leading digit survives.

## The fix

Prefix `__` in that case, matching xxd. After:

```
$ hexyl -i 9lives.bin | head -1
unsigned char __9lives_bin[] = {
$ hexyl -i 9lives.bin > z.c && gcc -c z.c    # exits 0
```

Names that do not start with a digit are untouched: `my-file.bin` still gives `my_file_bin`.

## One case I deliberately left alone

`is_alphanumeric` is Unicode-aware, so `té.bin` becomes `té_bin`. That looked wrong to me at first, but gcc accepts it (C99 universal character names in identifiers) and it compiles, so I left it rather than widen the diff. Say the word if you would rather restrict the whole thing to ASCII.

## Verification

`cargo test` is 5 and 41 passing with 0 failures, `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean.

A test alongside the existing `include_mode_from_file` covers the digit case. Reverting the guard with the test kept fails it on exactly the identifier.

Changelog entry added under a new unreleased section, per CONTRIBUTING. There was no unreleased section yet, so I created one; move it if you would rather it went elsewhere.

There is no issue for this, I found it while reading the include-mode code.


VXNCXNX:
This one has been open a month, is green and level with `master`, so it is only waiting on a look.

`+34/-1`, two files. `hexyl -i` on a file whose name starts with a digit emits C that does not compile, because a C identifier cannot begin with one:

```
$ hexyl -i 9lives.bin > z.c && gcc -c z.c
z.c:1:15: error: invalid suffix "lives_bin" on integer constant
    1 | unsigned char 9lives_bin[] = {
```

`xxd -i`, the tool #242 asked for parity with, prefixes instead of failing, and this does the same.

Happy to rebase or trim if it is more than you want here.

